PR #6, opened 2023-01-25
Added queueing and implemented delegate callbacks

TimEvens:


src/safe_queue.hpp
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) 2023 Cisco Systems, Inc. and others.  All rights reserved.
```
suhasHere, line 2: can we use .h to be consistent

src/safe_queue.hpp
```diff
@@ -0,0 +1,95 @@
...
+   * @brief inserts element at the end of queue
+   *
+   * @param elem
+   * @return True if successfully pushed, false if not.  The cause for false is
+   * that the queue is full.
+   */
+  bool push(T const &elem) {
+    if (limit and size() >= limit) {
+      return false;
+    }
```
suhasHere, line 43: s/and/&&

src/safe_queue.hpp
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * @brief Remove the first object from queue (oldest object)
+   *
+   * @return std::nullopt if queue is empty, otherwise reference to object
+   */
+  std::optional<T> pop() {
+    std::lock_guard<std::mutex> lock(mutex);
+    if (queue.empty()) {
+      return {};
+    }
```
suhasHere, line 59: return std::nulllopt

src/safe_queue.hpp
```diff
@@ -0,0 +1,95 @@
...
+template <typename T> class safeQueue {
+public:
+  /**
+   * @brief safeQueue constructor
+   *
+   * @param limit     Limit number of messages in queue before push blocks. Zero
+   *                  is unlimited.
+   */
+  safeQueue(uint32_t limit = 0) { this->limit = limit; }
+
```
suhasHere, line 30: limit = 0 as default is problematic .. may be we should have say video objects for 5 seconds worth of data

cmd/echoServer.cc
```diff
@@ -19,10 +19,46 @@ to_hex(const std::vector<uint8_t>& data)
 }
 
 struct Delegate : public ITransport::TransportDelegate {
-	void on_connection_status(const TransportContextId &context_id, const TransportStatus status) {}
-	void on_new_connection(const TransportContextId &context_id, const TransportRemote &remote) {}
-	void on_recv_notify(const TransportContextId &context_id) {}
-	void onNewMediaStream(const TransportContextId &context_id, const MediaStreamId &mStreamId) {}
+private:
+	std::shared_ptr<ITransport> *server;
+	uint64_t msgcount;
```
suhasHere, line 24: we don't need *server here

cmd/echoServer.cc
```diff
@@ -33,17 +69,12 @@ int main()
...
     uint64_t msid = 0; /* unused */
-    while (1)
-    {
 
-        auto data = server->dequeue(tcid, msid);
-        if (!data.has_value()) {
-            std::this_thread::sleep_for(std::chrono::milliseconds(1));
-            continue;
-        }
+		d.setServerTransport(&server);
```
suhasHere, line 73: we can just send the server as-is and take the shared_ptr everywhere

suhasHere (changes requested):
few comments ..but overall looking good .. thanks

PR #15, opened 2023-02-03
Resolves #12, updates logger docs, and improves cpu with fd_writer

### LINKED ISSUE #12: Send errors instead of asserts
There are several asserts on errors.  Change all to return error status/code and/or callback.

TimEvens:
fd_writer() used to use a sleep, which was inefficient and introduced additional delays. The new ```block_pop()``` method adds blocking to the queue if the queue is empty.  This improves CPU and removes the additional delay caused by the sleep.

Instead of using asserts, throws were added. The caller should catch these. Logging has been added.


TimEvens:
resolves #12

PR #23, opened 2023-02-25
Addressed some threading issues that prevented proper shutdown

paulej:
I was trying to add code to properly shut down the "really" server and I found threads were getting hung, mutexes were being accessed when they should not, logger reference was accessed when it no longer existed.

I addressed part of the issues here, part in forthcoming changes to "really".

I noted that references are being passed around and stored in objects.  For example:

```cpp
LogHandler& logger;
```

This is in transport_udp.h.  If this transport object is not destroyed before the parent, this causes a crash.  This is _exactly_ like passing around raw pointers.  We should be using shared pointers for such things like this.  That, or use weak pointers and check each before accessing.  Generally, the latter is no needed so long as one doesn't create circular dependencies.

### DIFF
--- a/src/safe_queue.h
+++ b/src/safe_queue.h
@@ -7,6 +7,7 @@
 #include <optional>
 #include <queue>
 #include <unistd.h>
+#include <condition_variable>
 
 namespace qtransport {
 
@@ -29,12 +30,12 @@ public:
    *                  is unlimited.
    */
   safeQueue(uint32_t limit = 1000)
+    : stop_waiting{ false }
+    , limit{ limit }
   {
-    this->limit = limit;
-    empty_block_mutex.try_lock();
   }
 
-  ~safeQueue() {}
+  ~safeQueue() { stopWaiting(); }
 
   /**
    * @brief inserts element at the end of queue
@@ -45,16 +46,15 @@ public:
    */
   bool push(T const& elem)
   {
+    std::lock_guard<std::mutex> lock(mutex);
+
     if (limit && size() >= limit) {
       return false;
     }
 
-    std::lock_guard<std::mutex> lock(mutex);
-
-    if (size() == 0)
-      empty_block_mutex.unlock();
-
     queue.push(elem);
+
+    cv.notify_one();
 
     return true;
   }
@@ -67,19 +67,8 @@ public:
   std::optional<T> pop()
   {
     std::lock_guard<std::mutex> lock(mutex);
-    if (queue.empty()) {
-      return std::nullopt;
-    }
 
-    auto elem = queue.front();
-    queue.pop();
-
-    // Lock empty lock to trigger blocking of threads using block_pop()
-    if (size() == 0) {
-      empty_block_mutex.try_lock();
-    }
-
-    return elem;
+    return popInternal();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/transport_udp.cpp
```diff
@@ -25,30 +25,30 @@ UDPTransport::~UDPTransport()
...
 
   // Close socket fd
-  ::close(fd);
+  if (fd >= 0)
+    ::close(fd);
 
   logger.log(LogLevel::info, "Closing transport threads");
-  for (auto thr : running_threads) {
-    thr->join();
+  for (auto &thread : running_threads) {
```
TimEvens, line 36: This for preference?

src/transport_udp.cpp
```diff
@@ -258,8 +258,8 @@ UDPTransport::fd_reader()
                         &remoteAddrLen);
 
     if (rLen < 0) {
-      if (errno == EAGAIN) {
-        // timeout on read
+      if ((errno == EAGAIN) || (stop)) {
```
TimEvens, line 261: I also ran into the reader thread not closing nicely due to FD being closed, but adding ```errno == EBADF``` sufficed.  This is needed anyway for use cases that the FD is closed by other means. Difference is that we don't want to ```continue```, instead ```break```.

TimEvens (approved):
Changing to CV instead of dedicated mutex is a net=0 change, but is more modern, so that's nice. I'm unclear though on your issues.  Changing to CV doesn't fix that, it is just a different way to handle the blocking.  fd > 0 vs fd >= 0 is a good call but not likely the issue you were running into.

It would be great if you can provide a bit more context on the issue you ran into. 

Regarding shutting down the server, such as **really**, I did run into abrupt terminations where logging would indicate error on exit. These were benign of course. The cause was not with the threading at all. It was with ```fd_reader()``` blocking on ```recvfrom()``` and closing the socket fd while waiting to read.  This caused **really** to exit with an error vs a nice logging message saying exiting normally.  The reason was because ```recvfrom()``` is blocking by design and when we close the FD, it interrupts the blocking read and generates a BAD FILE DESCRIPTOR error.  Your change does include a fix for this, but it suffers from race and other conditions still where the socket can be closed.  The change you added adds ``` || (stop)```  to the EAGAIN if statement.  This is good to add, but we still need to add ```errno == EBADF``` which will handle closing the FD at any point, controlled or uncontrolled.    Also, closing the FD and when we stop should ```break``` instead of ```continue```.

src/safe_queue.h
```diff
@@ -45,17 +46,16 @@ class safeQueue
...
     }
 
-    std::lock_guard<std::mutex> lock(mutex);
-
-    if (size() == 0)
-      empty_block_mutex.unlock();
-
     queue.push(elem);
 
+    cv.notify_one();
```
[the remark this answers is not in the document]
paulej: If there are waiting threads on that "pop block" function, then this will awaken them.  If no waiting threads, this has no real impact.

src/safe_queue.h
```diff
@@ -45,17 +46,16 @@ class safeQueue
    */
   bool push(T const& elem)
   {
+    std::lock_guard<std::mutex> lock(mutex);
```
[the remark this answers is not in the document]
paulej: size() is going to be a problem if called externally.  We should probably create a private version that assumes the mutex is locked, like I did the popInternal().

src/transport_udp.cpp
paulej: vectors are almost always preferred over lists and pointers are almost always never preferred.  A list of pointers is a bad thing to use.

src/transport_udp.cpp
paulej: I just put the check for stop() here so that processing would not continue on stop.  I didn't try to address all possible failure scenarios.  I do wonder why we use this as opposed to the Cantina::Network object.  It has proven to be pretty robust, as we're using that in Gallia.  It can use one or more threads to deliver data packets asynchronously to the application.  I also used it in QoH3.

PR #27, opened 2023-03-22
Simple install

shenning00:
Just added simple install for projects that don't want to use submodules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,4 +59,4 @@ endif()
 install(TARGETS ${PROJECT_NAME}
         LIBRARY DESTINATION lib)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -55,3 +55,8 @@ endif()
 if(BUILD_TESTING AND QTRANSPORT_BUILD_TESTS)
   add_subdirectory(cmd)
 endif()
+
+install(TARGETS ${PROJECT_NAME}
+        LIBRARY DESTINATION lib)
```
RichLogan, line 60: Could you give a `FRAMEWORK DESTINATION` too? Annoyingly it is not optional for CMake and causes an error when building to Apple frameworks.

RichLogan (changes requested):

TimEvens:
@shenning00 , can we close or update this?

PR #28, opened 2023-04-03
Quic transport

TimEvens:


src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+    switch (cb_mode) {
+      case picoquic_packet_loop_ready: {
+        log_msg << "packet_loop_ready, waiting for packets";
+        transport->logger.log(LogLevel::info, log_msg.str());
+
+        if (transport->isServerMode)
+          transport->setStatus(TransportStatus::Ready);
+
+        if (callback_arg != nullptr) {
+          auto *options = (picoquic_packet_loop_options_t *) callback_arg;
```
GhostofCookie, line 196: Should prefer explicit casting to c-style casting.

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+    case picoquic_callback_ready: { // Connection callback, not per stream
+      if (stream_cnx == NULL) {
+        stream_cnx = transport->createStreamContext(cnx, stream_id);
+      }
+
+      if (transport->isServerMode) {
+
+        // Create new stream context for new stream/connection
+        picoquic_enable_keep_alive(cnx, 3000000);
+        (void)picoquic_mark_datagram_ready(cnx, true);
```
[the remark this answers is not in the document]
GhostofCookie: Okay, I see this happening a lot further down as well. Are these returns not useful?

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+    cbNotifyThread.join();
+  }
+
+  logger.log(LogLevel::info, "done closing transport threads");
+
+}
+
+
+TransportStatus
+PicoQuicTransport::status()
```
GhostofCookie, line 424: Should this be a const member function?

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+
+  picoquic_set_default_tp(quic_ctx, &local_tp_options);
+
+  cbNotifyThread = std::thread(&PicoQuicTransport::cbNotifier, this);
+
+  if (isServerMode) {
+    picoQuicThread = std::thread(&PicoQuicTransport::server, this);
+    return 0;
+
+  } else {
```
GhostofCookie, line 515: Since we return in the if statement, else is redundant.

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+  // Set next stream ID
+  next_stream_id += 4; // Increment by 4, stream type is first 2 bits
+
+  return stream_cnx->stream_id;
+}
+
+TransportContextId
+PicoQuicTransport::start()
+{
+  TransportContextId cid = 0;
```
GhostofCookie, line 478: This isn't even used until the very end, why not declare it down there?

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+            sendOutData(&s_pair.second, NULL, 1300);
+          } else {
+            (void)picoquic_mark_active_stream(s_pair.second.cnx,
+                                              s_pair.first,
+                                              1,
+                                              (StreamContext*)&s_pair.second);
+          }
+      }
+    }
+  }
```
GhostofCookie, line 676: A lot of arrow code, and not using some of the nice stuff about about C++17/20 (i.e. structured-bindings). Makes it hard to read with all the `first`s and `second`s.

For example:
```cpp
for (auto& [_, active_stream] : active_streams) {
  for (auto& [ctx, data] : active_stream) {
    if (ctx == 0) {
      sendOutData(&data, NULL, 1300);
      continue;
    }
    // Awkward to format in GH editor so just imagine the params on newlines same as your code
    (void)picoquic_mark_active_stream(data.cnx, ctx, 1, reinterpret_cast<StreamContext*>(&data));
  }
}
```
Given it's a double for loop it is understandably hard to reduce the amount of tabbing, so arrow code is pretty much the only way.

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+                                    size_t max_len)
+{
+  for (int i=0; i < 120; i++) {
+    const auto& out_data = stream_cnx->out_data.pop();
+    if (out_data.has_value()) {
+//          std::ostringstream log_msg;
+//          log_msg << "Send Data: stream: " << stream_cnx->stream_id
+//                  << " max_len: " << max_len
+//                  << " queue_size: " << stream_cnx->out_data.size();
+//          logger.log(LogLevel::info, log_msg.str());
```
GhostofCookie, line 690: Seeing a lot of dead log messages, should we revive them and hide them behind some macro or flag?

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+      }
+    }
+  }
+}
+
+void PicoQuicTransport::sendOutData(StreamContext *stream_cnx,
+                                    [[maybe_unused]] uint8_t* bytes_ctx,
+                                    size_t max_len)
+{
+  for (int i=0; i < 120; i++) {
```
GhostofCookie, line 683: Why 120? (Curiousity question)

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+//          log_msg << "Send Data: stream: " << stream_cnx->stream_id
+//                  << " max_len: " << max_len
+//                  << " queue_size: " << stream_cnx->out_data.size();
+//          logger.log(LogLevel::info, log_msg.str());
+
+      if (stream_cnx->stream_id == 0 and max_len >= out_data.value().bytes.size() ) {
+          (void)picoquic_queue_datagram_frame(
+            stream_cnx->cnx, out_data->bytes.size(), out_data->bytes.data());
+
+      } else if (max_len > 1000) {
```
GhostofCookie, line 696: Why 1000?

src/transport_picoquic.cpp
```diff
@@ -0,0 +1,828 @@
...
+        }
+
+      } else {
+        return TransportError::InvalidStreamId;
+      }
+  } else {
+      return TransportError::InvalidContextId;
+  }
+  return TransportError::None;
+}
```
GhostofCookie, line 756: Can we error out early and use fewer tabs?
```cpp
if (ctx == active_streams.end()) {
  return TransportError::InvalidContextId;
}

const auto& stream_cnx = ctx->second.find(stream_id);
if (stream_cnx == ctx->second.end()) {
  return TransportError::InvalidStreamId;
}

if (!stream_cnx->second.out_data.push(out_data)) {
  return TransportError::QueueFull;
}

return TransportError::None;

File: src/transport_picoquic.cpp (hunk 6 above)
Comment (GhostofCookie) [reply src/transport_picoquic.cpp:676]: Just realized from reading other code further down I should've called `data` `stream` instead.

File: src/transport_picoquic.h [hunk 11]
```diff
@@ -0,0 +1,159 @@
...
+    const StreamId & stream_id) override;
+
+  /*
+   * Internal public methods
+   */
+  void setStatus(TransportStatus status);
+
+  StreamContext * getZeroStreamContext(picoquic_cnx_t* cnx);
+
+  StreamContext * createStreamContext(picoquic_cnx_t* cnx,
```
GhostofCookie, line 102: Why return raw pointers?

src/transport_udp.cpp
```diff
@@ -52,7 +52,7 @@ UDPTransport::UDPTransport(const TransportRemote& server,
 }
 
 TransportStatus
-UDPTransport::status() const
+UDPTransport::status()
```
GhostofCookie, line 55: Why not const?

src/transport_udp.cpp
```diff
@@ -295,19 +295,19 @@ UDPTransport::fd_reader()
...
+            last_stream_id,
         };
 
         cd.contextId = last_context_id;
-        cd.mStreamId = last_media_stream_id;
+        cd.streamId = last_stream_id;
 
         // Create dequeue
-        dequeue_data_map[last_context_id][last_media_stream_id].setLimit(50000);
+        dequeue_data_map[last_context_id][last_stream_id].setLimit(50000);
```
GhostofCookie, line 310: Should 50000 be a named constexpr at the top of this file or in the header somewhere?

src/transport_udp.h
```diff
@@ -61,24 +61,24 @@ class UDPTransport : public ITransport
 
   virtual ~UDPTransport();
 
-  TransportStatus status() const override;
+  TransportStatus status() override;
```
GhostofCookie, line 64: :(

GhostofCookie (approved):
Looks good, approving.

My comments are just pedantic modern C++ nonsense or just noob questions.

src/transport_picoquic.cpp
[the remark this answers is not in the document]
TimEvens: Not needed and not implemented right now.  Voiding them seems to address some other warnings.  We might want to add error handling later for some of these.

src/transport_picoquic.cpp
TimEvens: I can update them, but some of them require ```reinterpret_cast```.  It gets a bit more verbose, but it should be fine.

src/transport_picoquic.cpp
TimEvens: That should be fine. I'll update it. It does require updates to the interface since it's an implemented method.

src/transport_picoquic.cpp
TimEvens: IMO, it's easier to read with the else.. ```cid``` is initialized to ZERO, so it should be fine to remove the ```return 0```.

src/transport_picoquic.cpp
TimEvens: I'll move it to be more readable.

src/transport_udp.cpp
TimEvens: Reverted that.

src/transport_udp.cpp
```diff
@@ -20,7 +20,7 @@ using namespace qtransport;
 
 UDPTransport::~UDPTransport()
 {
-  // TODO: Close all media streams and connections
+  // TODO: Close all streams and connections
```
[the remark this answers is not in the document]
TimEvens: This is an older comment from before. It shows up because of the name change from "media stream" to "stream".

src/transport_picoquic.h
TimEvens: It's because of having to pass them to C functions (picoquic).

src/transport_udp.cpp
TimEvens: It should be a config variable, which we should add after this PR. There are several other config variables to add, which requires passing them upstream to the app so that they can set them.

GhostofCookie (approved):

PR #32, opened 2023-04-17
STT updates:

TimEvens:
* Rename in/out to rx/tx
* Rename safe queue methods to be more consistent naming
* Reduce default queue sizes as it was causing lag
* Add config param for rx/tx queue size


TimEvens:
more updates coming

shenning00 (commented):
Looking forward to the next round of changes.

shenning00 (approved):

PR #33, opened 2023-04-26
Reducing the amount of tests and executables being compiled.

GhostofCookie:
This PR itself doesn't actually reduce the building, just sets the flags. It depends on the modules picoquic and picotls being updated.

### DIFF
--- a/cmd/client.cc
+++ b/cmd/client.cc
@@ -12,15 +12,6 @@ using namespace qtransport;
 bool done = false;
 using bytes = std::vector<uint8_t>;
 
-static std::string to_hex(const std::vector<uint8_t> &data) {
-  std::stringstream hex(std::ios_base::out);
-  hex.flags(std::ios::hex);
-  for (const auto &byte : data) {
-    hex << std::setw(2) << std::setfill('0') << int(byte);
-  }
-  return hex.str();
-}
-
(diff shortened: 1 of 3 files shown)


cmd/client.cc
```diff
@@ -12,15 +12,6 @@ using namespace qtransport;
...
 
-static std::string to_hex(const std::vector<uint8_t> &data) {
-  std::stringstream hex(std::ios_base::out);
-  hex.flags(std::ios::hex);
-  for (const auto &byte : data) {
-    hex << std::setw(2) << std::setfill('0') << int(byte);
-  }
-  return hex.str();
-}
-
```
GhostofCookie, line 23: This was the only change (reducing a warning), the rest was the formatter.

cmd/echoServer.cc
```diff
@@ -8,15 +8,6 @@
...
 
-static std::string to_hex(const std::vector<uint8_t> &data) {
-  std::stringstream hex(std::ios_base::out);
-  hex.flags(std::ios::hex);
-  for (const auto &byte : data) {
-    hex << std::setw(2) << std::setfill('0') << int(byte);
-  }
-  return hex.str();
-}
-
```
GhostofCookie, line 19: Real change, the rest is formatter.

TimEvens (approved):
LGTM

PR #36, opened 2023-05-12
Use new picoquic API and cleanup queueing

TimEvens:


src/safe_queue.h
```diff
@@ -53,21 +52,19 @@ class safeQueue
   bool push(T const& elem)
   {
     bool rval = true;
+
     std::lock_guard<std::mutex> lock(mutex);
 
-    if (queue.size() >= limit) {// Make room by removing first element
+    if (queue.size() == 0)
```
GhostofCookie, line 58: Prefer `queue.empty()` for clarity, but that's a style choice.

src/safe_queue.h
```diff
@@ -178,14 +174,7 @@ class safeQueue
...
 
-    if (queue.size() < limit)
-      is_full = false;
-
-    if (queue.size() > 0) {
-      cv.notify_one();
-    }
-
-    return elem;
+    return std::move(elem);
```
GhostofCookie, line 177: Copy elision should handle this for you, I'm a little surprised the compiler didn't issue a warning about it (since explicitly calling move prevents elision).

GhostofCookie: Somre reading:
https://stackoverflow.com/questions/14856344/when-should-stdmove-be-used-on-a-function-return-value
https://en.cppreference.com/w/cpp/language/copy_elision

src/transport_picoquic.cpp
```diff
@@ -855,18 +815,27 @@ PicoQuicTransport::enqueue(const TransportContextId& context_id,
...
+
+        /*
+      if (stream_cnx->second.stream_id) {
+          (void) picoquic_mark_active_stream(stream_cnx->second.cnx,
+                                             stream_cnx->second.stream_id,
+                                             1,
+                                             static_cast<StreamContext *>(&stream_cnx->second));
+      } else {
+        (void) picoquic_mark_datagram_ready(stream_cnx->second.cnx, 1);
+      }*/
```
GhostofCookie, line 836: Is this needed in the future? We could grab it from source control if we need it again.

GhostofCookie (approved):

PR #38, opened 2023-06-06
Small memory leak fix.

shenning00:
getaddrinfo needs a corresponding freeaddrinfo.

I think I put this in the right place. 

Not sure if it needs a check first to see if the address list is null.


TimEvens (approved):
LGTM

PR #41, opened 2023-06-14
Priority queue

TimEvens:
Changes to use priority queue for transmit and time-queue for receive.

PR #42, opened 2023-06-15
Making StreamId a more concrete type.

GhostofCookie:


include/transport/stream_id.h
```diff
@@ -0,0 +1,96 @@
...
+
+    constexpr StreamId operator++(int) noexcept {
+      StreamId id(*this);
+      ++*this;
+      return id;
+    }
+
+  private:
+    constexpr void adjust(type value) {
+      _value = (value & (~0x0u << 2)) | _is_server | (_is_unidirectional * 2);
```
GhostofCookie, line 58: Not showing binary literals here with ternary because:
- `_is_server` can only ever be `0b00` or `0b01`, so use the value directly
- `_is_unidirectional` can only ever be `0b00` or `0b01`, so multiply by 2 to get either `0b00` or `0b10`

include/transport/stream_id.h
```diff
@@ -0,0 +1,98 @@
...
+    auto v = load();
+    store(++v);
+    return *this;
+  }
+
+  atomic operator++(int) {
+    auto v = load();
+    auto v_out = v++;
+    store(v);
+    return v_out;
```
GhostofCookie, line 94: This looks a little bulky and dumb, but it gets the job done without delving into the deep dark abyss that is system header c++ code.

include/transport/stream_id.h
suhasHere: are we creating QUIC StreamIds outside QUIC ?

suhasHere (changes requested):
I see code that is trying to recreate streamId definition form QUIC.. there are picoquic APIs for the same . May be I don't fully understand the context and hence missing something here ?

TimEvens:
While the changes are good, they will cause merge conflicts with #41. I recommend we merge #41 the merge and deal with these conflicts

include/transport/stream_id.h
GhostofCookie: Oops, that's a good point

PR #43, opened 2023-06-15
Removing double increment of next stream id.

GhostofCookie:


src/transport_picoquic.cpp
```diff
@@ -531,34 +514,25 @@ PicoQuicTransport::setStatus(TransportStatus status)
...
-    next_stream_id = ::make_stream_id(next_stream_id + 4, _is_server_mode, _is_bidirectional);
-
-    PicoQuicTransport::StreamContext *stream_cnx = createStreamContext(
-      cnx_stream_iter->second.cnx,
-      next_stream_id);
+        return datagram_stream_id;
 
-    TransportContextId tcid = context_id;
-    StreamId stream_id = next_stream_id;
+    next_stream_id += 4;
```
GhostofCookie, line 531: Just incrementing by 4 since we set the server/direction bits on construction. Of course, using `::make_stream_id` would protect against accidental increments of those bits but we're unlikely to do it here, so..... it's fine.

PR #47, opened 2023-07-11
Fix busy loop in picoquic callbacks

suhasHere:
PR to address busy loop callback fix. This needs to be verified with an end to end test once to make sure everything else is sane.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,6 @@ endif()
 ### Applications
 ###
 
-if(BUILD_TESTING AND QTRANSPORT_BUILD_TESTS)
+#if(BUILD_TESTING AND QTRANSPORT_BUILD_TESTS)
   add_subdirectory(cmd)
-endif()
+#endif()
--- a/cmd/client.cc
+++ b/cmd/client.cc
@@ -115,17 +115,32 @@ int main() {
   std::stringstream s_log;
 
   uint32_t *msg_num = (uint32_t *)&data_buf;
-
+  int num_messges = 1000;
+/**
   while (true) {
-    for (int i = 0; i < 12; i++) {
-      (*msg_num)++;
-      auto data = bytes(data_buf, data_buf + sizeof(data_buf));
-
-      client->enqueue(tcid, server.proto == TransportProtocol::UDP ? 1 : stream_id,
+    while(num_messges > 0) {
+        //std::cerr << "mesg: " << *msg_num << std::endl;
+        (*msg_num)++;
+        auto data = bytes(data_buf, data_buf + sizeof(data_buf));
+        client->enqueue(tcid, server.proto == TransportProtocol::UDP ? 1 : stream_id,
                         std::move(data));
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        num_messges -= 1;
     }
-    std::this_thread::sleep_for(std::chrono::milliseconds(20));
-  }
+
+    std::this_thread::sleep_for(std::chrono::seconds (5));
+    **/
+
+    while (true) {
+        for (int i = 0; i < 10; i++) {
+            (*msg_num)++;
+            auto data = bytes(data_buf, data_buf + sizeof(data_buf));
+
+            client->enqueue(tcid, server.proto == TransportProtocol::UDP ? 1 : stream_id,
+                            std::move(data));
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
 
   client->closeStream(tcid, stream_id);
 
--- a/cmd/client.cc
+++ b/cmd/client.cc
@@ -115,32 +115,18 @@ int main() {
   std::stringstream s_log;
 
   uint32_t *msg_num = (uint32_t *)&data_buf;
-  int num_messges = 1000;
-/**
+
+
   while (true) {
-    while(num_messges > 0) {
-        //std::cerr << "mesg: " << *msg_num << std::endl;
+    for (int i = 0; i < 10; i++) {
         (*msg_num)++;
         auto data = bytes(data_buf, data_buf + sizeof(data_buf));
+
         client->enqueue(tcid, server.proto == TransportProtocol::UDP ? 1 : stream_id,
                         std::move(data));
-        std::this_thread::sleep_for(std::chrono::milliseconds(10));
-        num_messges -= 1;
     }
-
-    std::this_thread::sleep_for(std::chrono::seconds (5));
-    **/
-
-    while (true) {
-        for (int i = 0; i < 10; i++) {
-            (*msg_num)++;
-            auto data = bytes(data_buf, data_buf + sizeof(data_buf));
-
-            client->enqueue(tcid, server.proto == TransportProtocol::UDP ? 1 : stream_id,
-                            std::move(data));
-        }
-        std::this_thread::sleep_for(std::chrono::milliseconds(10));
-    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  }
 
   client->closeStream(tcid, stream_id);
 
--- a/include/transport/priority_queue.h
+++ b/include/transport/priority_queue.h
@@ -99,7 +99,6 @@ namespace qtransport {
             auto& queue = _queue[priority];
             queue->push(value, ttl);
             //std::cerr << "added object " << num_objects << std::endl;
-            num_objects++;
         }
 
         /**
@@ -155,8 +154,6 @@ namespace qtransport {
                 if (_queue[i]) {
                     const auto& obj = _queue[i]->pop();
                     if (obj.has_value()) {
-                        //std::cerr << "removed object " << num_objects << std::endl;
-                        num_objects--;
                         return;
                     }
                 }
@@ -181,7 +178,6 @@ namespace qtransport {
         size_t _initial_queue_size;
         size_t _duration_ms;
         size_t _interval_ms;
-        uint64_t num_objects {0};
 
         std::array<std::unique_ptr<timeQueue>, PMAX> _queue;
 
--- a/include/transport/priority_queue.h
+++ b/include/transport/priority_queue.h
@@ -81,7 +81,7 @@ namespace qtransport {
      * @param value     The value to push onto the queue.
      * @param ttl       The time to live of the value in milliseconds.
      * @param priority  The priority of the value (range is 0 - PMAX)
-         */
+     */
         void push(DataType& value, uint32_t ttl, uint8_t priority = 0)
         {
             std::lock_guard<std::mutex> lock(_mutex);
@@ -98,6 +98,8 @@ namespace qtransport {
 
             auto& queue = _queue[priority];
             queue->push(value, ttl);
+            //std::cerr << "added object " << num_objects << std::endl;
+            num_objects++;
         }
 
         /**
@@ -153,6 +155,8 @@ namespace qtransport {
                 if (_queue[i]) {
                     const auto& obj = _queue[i]->pop();
                     if (obj.has_value()) {
+                        //std::cerr << "removed object " << num_objects << std::endl;
+                        num_objects--;
                         return;
                     }
                 }
@@ -160,7 +164,16 @@ namespace qtransport {
         }
 
         // TODO: Consider changing empty/size to look at timeQueue sizes - maybe support blocking pops
-        size_t size() const { return _queue.size(); }
+        size_t size() const {
+            uint64_t count = 0;
+            for (size_t i = 0; i < _queue.size(); i++) {
+                if (_queue[i]) {
+                    count += _queue[i]->size();
+                }
+            }
+            return count;
+        }
+
         bool empty() const { return _queue.empty(); }
 
       private:
@@ -168,6 +181,7 @@ namespace qtransport {
         size_t _initial_queue_size;
         size_t _duration_ms;
         size_t _interval_ms;
+        uint64_t num_objects {0};
 
         std::array<std::unique_ptr<timeQueue>, PMAX> _queue;
 
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -502,6 +502,17 @@ PicoQuicTransport::shutdown()
     picoquic_config_clear(&config);
 }
 
+void
+PicoQuicTransport::update_datagram_ready_status(picoquic_cnx_t *cnx, bool value) {
+    if(cnx == nullptr) {
+        return;
+    }
+
+    if (pq_datagram_ready != value) {
+        picoquic_mark_datagram_ready(cnx, value);
+    }
+}
+
 TransportStatus
 PicoQuicTransport::status() const
 {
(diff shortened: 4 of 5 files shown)


src/transport_picoquic.cpp
```diff
@@ -766,26 +781,30 @@ PicoQuicTransport::sendTxData(StreamContext* stream_cnx, [[maybe_unused]] uint8_
...
                 std::memcpy(buf, out_data.value().data(), out_data.value().size());
-
                 stream_cnx->tx_data->pop();
             }
         }
+    } else {
+        // has_value false
+        //std::cout << "send: picoquic_provide_datagram_buffer_ex masrking inactive" << std::endl;
+
+        picoquic_provide_datagram_buffer_ex(bytes_ctx, 0, 0);
```
GhostofCookie, line 807: Instead of adding an else clause, why not return early if there's no value? That would reduce the amount of tabbing in the body of the function, and remove the need for the comment inside this statement.

src/transport_picoquic.cpp
```diff
@@ -808,8 +827,10 @@ PicoQuicTransport::enqueue(const TransportContextId& context_id,
 
         if (stream_cnx != ctx->second.end()) {
             metrics.enqueued_objs++;
+            // add the app data to the queue
+            // mark the stream active if it isn't already
```
GhostofCookie, line 831: Seems redundant to state, the code below seems well written and self explanatory.

src/transport_picoquic.cpp
```diff
@@ -766,26 +781,30 @@ PicoQuicTransport::sendTxData(StreamContext* stream_cnx, [[maybe_unused]] uint8_
...
-            if (buf != NULL) {
+            if (buf != nullptr) {
                 std::memcpy(buf, out_data.value().data(), out_data.value().size());
-
                 stream_cnx->tx_data->pop();
             }
         }
+    } else {
+        // has_value false
+        //std::cout << "send: picoquic_provide_datagram_buffer_ex masrking inactive" << std::endl;
```
GhostofCookie, line 805: Dead code alert.

Do we still want this? We can probably write it again if we need it.

src/transport_picoquic.cpp
```diff
@@ -756,7 +771,7 @@ PicoQuicTransport::checkTxData()
 void
 PicoQuicTransport::sendTxData(StreamContext* stream_cnx, [[maybe_unused]] uint8_t* bytes_ctx, size_t max_len)
 {
-    if (bytes_ctx == NULL) {
+    if (bytes_ctx == nullptr) {
```
GhostofCookie, line 774:
```suggestion
    if (!bytes_ctx) {
```
Similar suggestion throughout, but this is a personal preference. Just feels weird to see NULL/nullptr everywhere.

Feel free to ignore this kind of comment :)

src/transport_picoquic.cpp
```diff
@@ -730,23 +741,27 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
-                if (s_pair.second.tx_data->size() > 0) {
+                // QUIC Datagram
+                if (stream.tx_data->size() > 0) {
                     // reinsert by wake is one way to get picoquic to callback sooner
                     // picoquic_reinsert_by_wake_time(s_pair.second.cnx->quic, s_pair.second.cnx, cur_time);
                     // Only send tx data here when using picoquic queueing, otherwise let the callbacks handle it
                     // sendTxData(&s_pair.second, NULL, 1400);
-
-                    (void)picoquic_mark_datagram_ready(s_pair.second.cnx, 1);
+                    //picoquic_mark_datagram_ready(stream.cnx, 1);
```
GhostofCookie, line 760: Dead code alert.

src/transport_picoquic.cpp
```diff
@@ -730,23 +741,27 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
+                    update_datagram_ready_status(stream.cnx, true);
                 } else {
-                    (void)picoquic_mark_datagram_ready(s_pair.second.cnx, 1);
+                    //picoquic_mark_datagram_ready(stream.cnx, 0);
+                    update_datagram_ready_status(stream.cnx, false);
```
GhostofCookie, line 764: Seems like both the if and else statement are practically identical? Maybe instead:
```suggestion
update_datagram_ready_status(stream.cnx, stream.tx_data->size() > 0);
```
The comment inside the if block could remain to provide context.

src/transport_picoquic.cpp
```diff
@@ -730,23 +741,27 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
 void
 PicoQuicTransport::checkTxData()
 {
-    for (auto& c_pair : active_streams) {
-        for (auto& s_pair : active_streams[c_pair.first]) {
+    // std::map<TransportContextId, std::map<StreamId, StreamContext>> active_streams;
```
GhostofCookie, line 744: This is in the header now right?

GhostofCookie (commented):
Code looks good, but I'll hold off approval until we can test it out a bit. Can try that today.

Of course, as long as the calls still go through the same, regardless of performance, then we could just YOLO merge it and then update AWS laps and Testflight and see how it performs.

src/transport_picoquic.cpp
suhasHere: we can return early since we need to let the stack know .. that was the one of the reasons for busy loop earlier

src/transport_picoquic.cpp
suhasHere: In the new changes if and else are doing opposite logic. Yes, the old code was doing exactly the same thing and causing more busy loop calls and new change should address it

src/transport_picoquic.cpp
suhasHere: yes, the comment will be removed

src/transport_picoquic.h
```diff
@@ -132,14 +133,17 @@ class PicoQuicTransport : public ITransport
...
     /*
    * Internal Public Variables
      */
     LogHandler& logger;
     bool _is_server_mode;
     bool _is_unidirectional{ false };
     bool debug {false};
-
+    // current status of our knowledge of picoquic datagram ready status set
+    std::atomic<bool> pq_datagram_ready { false };
```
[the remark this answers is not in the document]
suhasHere: it is prefixed with pq since it is picoquic api status .. also we at some point  in Jan agreed not to use _XXX and mXXX form of naming .. again i don't care either way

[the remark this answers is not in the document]
GhostofCookie: > not to use _XXX and mXXX form of naming .. again i don't care either way

:(

src/transport_picoquic.cpp
TimEvens: +1 for ```update_datagram_ready_status(stream.cnx, stream.tx_data->size() > 0);```

src/transport_picoquic.cpp
```diff
@@ -808,8 +827,10 @@ PicoQuicTransport::enqueue(const TransportContextId& context_id,
 
         if (stream_cnx != ctx->second.end()) {
             metrics.enqueued_objs++;
+            // add the app data to the queue
+            // mark the stream active if it isn't already
             stream_cnx->second.tx_data->push(bytes, ttl_ms, priority);
-
+            update_datagram_ready_status(stream_cnx->second.cnx, true);
```
TimEvens, line 833: If this change worked, it would mitigate the need for ```checkTxData()```. In other words, we would not need ```checkTxData()```. 

Unfortunately, this change beaks thread safety as the picoquic thread is not the same thread that enqueue is being run. Calling picoquic functions without locking. 

I tried running this branch and I do see the client consistently locking up after a little while, which is likely caused by this change.

TimEvens (changes requested):
`enqueue()` cannot call any picoquic functions without causing picoquic to crash or lockup doe to trie/list corruption.  The busy loop problem is caused by the fact that we cannot call these methods safely from another thread without having to add synchronization to picoquic.  Adding locking to picoquic could result in picoquic having stalls that result in failed ACKs or other QUIC level processing.

src/transport_picoquic.cpp
suhasHere: @TimEvens  very good point. not sure how i forgot about the single threaded nature of picoquic .. Let me update the fix

GhostofCookie: Can confirm the locking on laps from testing today with Scott. I will say, when used on client it didn't deadlock, and the CPU usage was noticeably reduced.

suhasHere: @GhostofCookie / @TimEvens how do i verify the locking issue or perhaps can you share where and how did you observe the locking issue. thanks

TimEvens: @suhasHere , it's a race condition, so you have to catch it.  The way I caught it was to increase the rate the client transmits data. 

```diff
diff --git a/cmd/client.cc b/cmd/client.cc
index 893c3f4..3c96d9a 100644
--- a/cmd/client.cc
+++ b/cmd/client.cc
@@ -117,14 +117,14 @@ int main() {
   uint32_t *msg_num = (uint32_t *)&data_buf;

   while (true) {
-    for (int i = 0; i < 12; i++) {
+    for (int i = 0; i < 10; i++) {
       (*msg_num)++;
       auto data = bytes(data_buf, data_buf + sizeof(data_buf));

       client->enqueue(tcid, server.proto == TransportProtocol::UDP ? 1 : stream_id,
                         std::move(data));
     }
-    std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
   }
```

TimEvens:
@suhasHere , let's close this as https://github.com/Quicr/transport/tree/relay-peering has a CPU fix that is working with picoquic.

suhasHere:
> @suhasHere , let's close this as https://github.com/Quicr/transport/tree/relay-peering has a CPU fix that is working with picoquic.

@TimEvens i have updated the PR with fixes . I want to see if we can avoid locking in picoquic. let's discuss the optiosn live

TimEvens (changes requested):
let's close this one for now.

suhasHere:
@TimEvens  I am not comfortable with closing this . I don't see why do you think this needs to be closed ?

CMakeLists.txt
```diff
@@ -53,6 +53,6 @@ endif()
 ### Applications
 ###
 
-if(BUILD_TESTING AND QTRANSPORT_BUILD_TESTS)
+#if(BUILD_TESTING AND QTRANSPORT_BUILD_TESTS)
   add_subdirectory(cmd)
-endif()
+#endif()
```
GhostofCookie, line 58: Would prefer we not build what we don't need in other projects where this is included.

cmd/client.cc
```diff
@@ -116,15 +116,16 @@ int main() {
 
   uint32_t *msg_num = (uint32_t *)&data_buf;
 
+
   while (true) {
-    for (int i = 0; i < 12; i++) {
-      (*msg_num)++;
-      auto data = bytes(data_buf, data_buf + sizeof(data_buf));
+    for (int i = 0; i < 10; i++) {
```
GhostofCookie, line 121: This 10 seems magic. I assume it has some relation to the sleep, but it would be nice if this had a name.

include/transport/priority_queue.h
```diff
@@ -98,6 +98,7 @@ namespace qtransport {
 
             auto& queue = _queue[priority];
             queue->push(value, ttl);
+            //std::cerr << "added object " << num_objects << std::endl;
```
GhostofCookie, line 101: DOA

include/transport/priority_queue.h
```diff
@@ -160,7 +161,16 @@ namespace qtransport {
...
 
         // TODO: Consider changing empty/size to look at timeQueue sizes - maybe support blocking pops
-        size_t size() const { return _queue.size(); }
+        size_t size() const {
+            uint64_t count = 0;
+            for (size_t i = 0; i < _queue.size(); i++) {
+                if (_queue[i]) {
+                    count += _queue[i]->size();
+                }
+            }
```
GhostofCookie, line 170: Could we use `std::accumulate` instead?

src/transport_picoquic.cpp
```diff
@@ -222,11 +222,11 @@ pq_event_cb(picoquic_cnx_t* cnx,
 int
 pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* callback_ctx, void* callback_arg)
 {
-
     PicoQuicTransport* transport = static_cast<PicoQuicTransport*>(callback_ctx);
     int ret = 0;
     std::ostringstream log_msg;
 
+    // call happens right before select
```
GhostofCookie, line 229: What "select"?

src/transport_picoquic.cpp
```diff
@@ -258,6 +258,7 @@ pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* ca
                 //        transport->logger.log(LogLevel::debug, log_msg.str());
                 break;
 
+
```
GhostofCookie, line 261: ?

src/transport_picoquic.cpp
```diff
@@ -269,16 +270,20 @@ pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* ca
...
+                    targ->delta_t = 10000;
+                } else {
+                    if(targ->delta_t > 10000) {
+                        // ensure we send the queued data immediately
+                        if (transport->checkTxData()) {
+                            targ->delta_t = 0;
+                            transport->update_datagram_ready_status(picoquic_get_first_cnx(quic), true);
+                        } else {
+                            targ->delta_t = 10000;
+                            transport->update_datagram_ready_status(picoquic_get_first_cnx(quic), false);
```
GhostofCookie, line 284: Magic `10000` copy pasted a bunch, these should be a constant somewhere.

GhostofCookie, line 284:
```suggestion
const bool enoughDelta = targ->delta_t > 10000;
targ->delta_t = enoughDelta ? 0 : 10000;
transport->update_datagram_ready_status(picoquic_get_first_cnx(quic), enoughDelta && transport->checkTxData());    
```
Obviously `enoughDelta` is a bad name, but I also have no idea what `targ->delta_t > 10000` means, so....

src/transport_picoquic.cpp
```diff
@@ -738,73 +739,62 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
     }
+
+    return available;
 }
 
 void
 PicoQuicTransport::sendTxData(StreamContext* stream_cnx, [[maybe_unused]] uint8_t* bytes_ctx, size_t max_len)
 {
     if (bytes_ctx == nullptr) {
+        //std::cerr << "N;";
```
GhostofCookie, line 771: DOA

src/transport_picoquic.cpp
```diff
@@ -738,73 +739,62 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
         if (max_len >= out_data.value().size()) {
             metrics.dgram_sent++;
-
             uint8_t* buf = nullptr;
-
             if (stream_cnx->stream_id == 0) {
                 buf = picoquic_provide_datagram_buffer_ex(bytes_ctx,
                                                           out_data.value().size(),
-                                                          1);
+                                                          (stream_cnx->tx_data->size() - 1) > 0);
```
GhostofCookie, line 785: I assume this is safe because size couldn't be 0 if we're in this `if` block, right?

src/transport_picoquic.cpp
```diff
@@ -738,73 +739,62 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
 void
 PicoQuicTransport::sendTxData(StreamContext* stream_cnx, [[maybe_unused]] uint8_t* bytes_ctx, size_t max_len)
 {
     if (bytes_ctx == nullptr) {
+        //std::cerr << "N;";
         metrics.send_null_bytes_ctx++;
         return;
     }
 
+
```
GhostofCookie, line 776: More double line breaks?

src/transport_picoquic.cpp
GhostofCookie: Or what `delta_t` is, but that looks like it came from somewhere else.... the _t is confusing because in C/C++ that typically means "_type", but I am guessing they are going for "_time"?

src/transport_picoquic.cpp
GhostofCookie: I see this survived?

src/transport_picoquic.cpp
```diff
@@ -738,73 +739,62 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
-                    // picoquic_reinsert_by_wake_time(s_pair.second.cnx->quic, s_pair.second.cnx, cur_time);
-                    // Only send tx data here when using picoquic queueing, otherwise let the callbacks handle it
-                    // sendTxData(&s_pair.second, NULL, 1400);
-                    //picoquic_mark_datagram_ready(stream.cnx, 1);
-                    update_datagram_ready_status(stream.cnx, true);
-                } else {
-                    //picoquic_mark_datagram_ready(stream.cnx, 0);
-                    update_datagram_ready_status(stream.cnx, false);
+                    available = true;
+                    break;
```
GhostofCookie, line 758: Are there no other streams than 0? Seems unlikely...

src/transport_picoquic.cpp
```diff
@@ -738,73 +739,62 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
-
             if (sid != 0) {
-                // QUIC Stream
-                if (stream.tx_data->size() > 0)
-                    (void)picoquic_mark_active_stream(
-                      stream.cnx, sid, 1, static_cast<StreamContext*>(&stream));
+                if (stream.tx_data->size() > 0) {
+                    (void) picoquic_mark_active_stream(
+                            stream.cnx, sid, 1, static_cast<StreamContext *>(&stream));
+                }
```
GhostofCookie, line 753: So, if there is no `sid==0`, then there is no available data?

GhostofCookie (changes requested):

TimEvens:
Close for now.  Other changes were added.

PR #49, opened 2023-07-21
Quieting time-queue bucket emptying.

GhostofCookie:
This looks crazier than it is, the formatter got at it.

Nothing has actually changed functionality-wise in time-queue, it just was really really loud, and wiping out logs. As well, it's probably not it's job to do metrics/logging in there, so removed TODOs around that.


TimEvens (approved):
LGTM

PR #50, opened 2023-07-23
Relay peering

TimEvens:
This fix removes some debug messages that were annoying, fixes picoquic CPU usage, and adds peering updates. Picoquic is also updated to latest master branch.


suhasHere:
@TimEvens let's check on the reduction of cpu usage based on #47 once as well. Then we can pick one and go with it.

TimEvens:
> @TimEvens let's check on the reduction of cpu usage based on #47 once as well. Then we can pick one and go with it.

Based on testing both branches, this PR directly solves the problem with picoquic and reduces CPU by 10% more than PR#47.  PR#47 doesn't directly solve the issue.  Regardless, this is working and it has better CPU reduction.

TimEvens:
@suhasHere , this PR also adds relay peering, fixes some logs/debugs that were not needed, and updates picoquic to the latest master branch.  We need this PR regardless.

suhasHere (dismissed):
I would like yo hold off picoquic locking changes until we verify #47 doesn't work or we do some testing on either of the approaches

TimEvens:
> I would like yo hold off picoquic locking changes until we verify #47 doesn't work or we do some testing on either of the approaches

@suhasHere, so the ball is in your court then. You are currently blocking relay-peering. Make sure to update this branch with whatever.

suhasHere:
i see relay-peering and picoquic are 2 parallel efforts. May be should we move picoquic investigations to #47 altogether ? thoughts ?

TimEvens:
@GhostofCookie , @shenning00 ,

I've removed mutex from picoquic and moved to use `picoquic_runner_queue` because of issues with stream creation from other threads. We need to run several picoquic functions in a thread safe way, not just mark datagram.  This change allows us to do that.  The picoquic event loop via the loop callback will now run the functions based on the other thread requests.

src/transport_picoquic.cpp
```diff
@@ -592,6 +594,18 @@ PicoQuicTransport::start()
...
+void PicoQuicTransport::pq_runner() {
+
+    while (picoquic_runner_queue.size() > 0) {
+        auto cb = std::move(picoquic_runner_queue.pop());
+        if (cb) {
+            (*cb)();
+        } else {
+            break;
+        }
+    }
```
GhostofCookie, line 606: Could this be achieved the same way by doing:
```cpp
do {
    if(auto cb = std::move(picoquic_runner_queue.pop())) (*cb)();
} while (picoquic_runner_queue.size() > 0);
```
?

src/transport_picoquic.cpp
```diff
@@ -810,12 +799,24 @@ PicoQuicTransport::enqueue(const TransportContextId& context_id,
             metrics.enqueued_objs++;
             stream_cnx->second.tx_data->push(bytes, ttl_ms, priority);
 
+            if (!stream_id) {
+                picoquic_runner_queue.push([=, this]() {
```
GhostofCookie, line 803: I see this for everything pushed to the runner queue: Do we want to capture by value? That does mean a copy of everything used (except for `this`) every time we push to the queue.

src/transport_picoquic.cpp
```diff
@@ -810,12 +799,24 @@ PicoQuicTransport::enqueue(const TransportContextId& context_id,
...
+                });
+            }
         } else {
             std::cerr << "enqueue dropped due to invalid stream: " << stream_id << std::endl;
             return TransportError::InvalidStreamId;
         }
     } else {
-        std::cerr << "enqueue dropped due to invalid contextId: " << context_id << std::endl;
+        //std::cerr << "enqueue dropped due to invalid contextId: " << context_id << std::endl;
+        //TODO: Add metrics to report invalid context
```
GhostofCookie, line 819: This has been spotted a lot recently, maybe we could keep this log around? Unless it isn't helpful, then never mind.

src/transport_picoquic.cpp
GhostofCookie: Or, even just omitting the break, since if it is null then nothing will happen, then it will try to loop, hit the condition, and then break the loop anyways.

TimEvens: Yep, I can update it.

src/transport_picoquic.cpp
TimEvens: Good catch. It was a cut/paste in. I'll update that.

src/transport_picoquic.cpp
TimEvens: Maybe it would be better to add it as a debug instead. When it does happen, it happens a lot and that becomes a lot of noise.  Normally we shouldn't see this though.

cmd/client.cc
```diff
@@ -117,14 +117,14 @@ int main() {
   uint32_t *msg_num = (uint32_t *)&data_buf;
 
   while (true) {
-    for (int i = 0; i < 12; i++) {
+    for (int i = 0; i < 10; i++) {
```
GhostofCookie, line 120: Bit of a magic number? Would be nice if it had a name.

include/transport/priority_queue.h
```diff
@@ -160,8 +159,20 @@ namespace qtransport {
...
+        size_t size() const {
+            size_t sz=0;
+            for (size_t i = 0; i < _queue.size(); i++) {
+                if (_queue[i]) {
+                    sz += _queue[i]->size();
+                }
+            }
+
+            return sz;
+        }
```
GhostofCookie, line 171: Could we use `std::accumulate`?

src/transport_picoquic.cpp
```diff
@@ -269,16 +270,10 @@ pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* ca
...
-                    << targ->current_time << " delta: " << targ->delta_t;
-                transport->logger.log(LogLevel::info, log_msg.str());
-                */
-
                 // TODO: Add config to set this value. This will change the loop select
                 //   wait time to delta value in microseconds. Default is <= 10 seconds
-                if (targ->delta_t > 1000) {
-                    targ->delta_t = 1000;
+                if (targ->delta_t > 5000) {
+                    targ->delta_t = 5000;
```
GhostofCookie, line 276: Magic numbers again. Also, I asked this in a different PR as well, but what is `delta_t`?

GhostofCookie: Oh, I see the comment above now, it looked like part of the TODO. "delta value in microseconds". Delta of... since we last waited?

src/transport_picoquic.cpp
```diff
@@ -544,7 +544,7 @@ TransportContextId
 PicoQuicTransport::start()
 {
     uint64_t current_time = picoquic_current_time();
-    debug_set_stream(stdout); // Enable picoquic debug
+    //debug_set_stream(stdout); // Enable picoquic debug
```
GhostofCookie, line 547: Should we add some ifdef around this in case we want it in debug mode? Or is it just noisy?

I'm all for quieting the library though.

include/transport/priority_queue.h
```diff
@@ -171,7 +171,13 @@ namespace qtransport {
...
         bool empty() const {
-            return size() == 0;
+            for (size_t i = 0; i < _queue.size(); i++) {
+                if (_queue[i] && !_queue[i]->empty()) {
+                    return false;
+                }
+            }
+
+            return true;
         }
```
GhostofCookie: Alternatively, we could let STL do it for us:
```suggestion
        bool empty() const {
            return std::all_of(_queue.begin(), _queue.end(), [&](const auto& elem) { return elem->empty(); });
        }
```

src/transport_picoquic.cpp
```diff
@@ -800,12 +796,12 @@ PicoQuicTransport::enqueue(const TransportContextId& context_id,
             stream_cnx->second.tx_data->push(bytes, ttl_ms, priority);
 
             if (!stream_id) {
-                picoquic_runner_queue.push([=, this]() {
+                picoquic_runner_queue.push([=]() {
```
GhostofCookie, line 799: Not quite what I meant, but I suppose performance isn't going to be lost here.

src/transport_picoquic.cpp
```diff
@@ -596,14 +596,10 @@ PicoQuicTransport::start()
...
-        auto cb = std::move(picoquic_runner_queue.pop());
-        if (cb) {
+    do {
+        if(auto cb = std::move(picoquic_runner_queue.pop()))
             (*cb)();
-        } else {
-            break;
-        }
-    }
+    } while (picoquic_runner_queue.size() > 0);
```
GhostofCookie: My second comment may have been more sane, I don't know why do/while came to mind first. This also just occurred to me as another valid way of doing this with fewer branches (and actually, may be what the compiler does under the hood, who knows):
```cpp
while (auto cb = std::move(picoquic_runner_queue.pop())) {
    (*cb)();
}

Comment (GhostofCookie) [reply src/transport_picoquic.cpp]: Is the move causing performance issues? Is it hiding copy elision?

File: include/transport/priority_queue.h [hunk 11]
```diff
@@ -160,8 +159,26 @@ namespace qtransport {
...
+
+        bool empty() const {
+            for (size_t i = 0; i < _queue.size(); i++) {
+                if (_queue[i] && !_queue[i]->empty()) {
+                    return false;
+                }
+            }
+
+            return true;
+        }
```
GhostofCookie, line 181: Alternatively, we could let STL do it for us:
```suggestion
        bool empty() const {
            return std::all_of(_queue.begin(), _queue.end(), [&](const auto& elem) { return elem->empty(); });
        }
```

GhostofCookie: Specifically from [std::all_of](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of):
> 1-2) true if unary predicate returns true for all elements in the range, false otherwise. Returns true if the range is empty.

cmd/client.cc
TimEvens: I'll add a config/env var for it. It's basically what you want to use for testing.  Same thing for the sleep_for.

src/transport_picoquic.cpp
TimEvens: It's the FD select delay in the picoquic loop.  It defaults to be a calculation from last time/wake but we are overriding it by setting it to a specific value if it's over it. If we didn't set this, it would seriously delay things when there isn't data being read. Need to add config for this too.  It's a reasonable number, but we might want to change it per client/server.

src/transport_picoquic.cpp
TimEvens: It wasn't noisy till recently with the latest picoquic master merge.  It's now pretty noisy. We can add a config option to enable or disable picoquic debug output.

src/transport_picoquic.cpp
TimEvens: Doesn't seem to make a diff on perf, but it's nicer to reduce lines.

GhostofCookie (approved):

PR #52, opened 2023-08-01
Fix both udp and picoquic termination

TimEvens:
* Picoquic transport issue was related to the picoquic callback not processing the return value. Fix is to add to the pq_loop_cb.
* UDP transport didn't have a receive timeout, which caused it to block forever until something was received. Even when something was received, it could result in a SEGV after the recvfrom when it was supposed to be stopped.


GhostofCookie (approved):
Code looks good!

PR #53, opened 2023-08-04
Code to retrieve the remove address for a transport context

paulej:
This will allow one to query the transport (UDP or QUIC) to get the peer's IP address.


src/transport_picoquic.cpp
```diff
@@ -739,6 +740,27 @@ PicoQuicTransport::closeStream(const TransportContextId& context_id, const Strea
     deleteStreamContext(context_id, stream_id);
 }
 
+bool PicoQuicTransport::getPeerAddrInfo(const TransportContextId& context_id,
+                                        sockaddr_storage* addr)
+{
+    // Locate the specified transport context
```
TimEvens, line 746: Not here, but we need to add locking to this post merge and merge of the streaming changes that adds the lock to use.

TimEvens (approved):
LGTM

src/transport_picoquic.cpp
paulej: Locking is needed in several places, as maps are consulted via several API calls.

PR #55, opened 2023-08-15
Fixing capture of stream context to close in pq

GhostofCookie:
Capturing the iterator was allowing the context (which was held by reference in the copied iterator) to be deleted before picoquic had a chance to properly close it, causing a crash. Reordered the if blocks a bit and added some named vars to replace the `iterator->second` calls.


TimEvens (approved):
LGTM

PR #59, opened 2023-08-24
Adjustments to time-queue to better handle latency spikes

TimEvens:


include/transport/safe_queue.h
```diff
@@ -143,6 +143,13 @@ class safeQueue
   }
 
   /**
+   * @brief Check if queue is empty
+   *
+   * @returns True if empty, false if not
+   */
+  bool empty() const { return size() > 0; }
+
```
suhasHere, line 151: If i am reading it correctly, it returns True if not empty ?
Suggest replacing 
```
return size() == 0;

```

include/transport/time_queue.h
```diff
@@ -232,16 +233,22 @@ namespace qtransport {
...
          * @throws std::invalid_argument    If the duration or interval do not meet requirements.
          * @throws std::runtime_error       If the timer is null.
          */
         time_queue(size_t duration,
                    size_t interval,
                    std::shared_ptr<queue_timer_service> timer,
-                   size_t initial_queue_size)
-          : time_queue(duration, interval, timer)
+                   size_t initial_queue_size,
+                   size_t spike_period_interval=30000,
```
suhasHere, line 246: Could we update the comment to specify the units of time here .. Is it 30 milliseconds ?

include/transport/transport.h
```diff
@@ -74,7 +74,7 @@ struct TransportConfig
   const uint32_t time_queue_init_queue_size {1000};     /// Initial queue size to reserve upfront
   const uint32_t time_queue_max_duration {1000};        /// Max duration for the time queue in milliseconds
   const uint32_t time_queue_bucket_interval {1};        /// The bucket interval in milliseconds
-  const uint32_t time_queue_rx_ttl { 250 };             /// Receive queue TTL
+  const uint32_t time_queue_rx_ttl { 500 };             /// Receive queue TTL
```
suhasHere, line 77: same comment as above, please add time units

src/transport_picoquic.cpp
```diff
@@ -1031,15 +1038,13 @@ PicoQuicTransport::on_recv_datagram(StreamContext* stream_cnx, uint8_t* bytes, s
...
-
     if (cbNotifyQueue.size() > 200) {
         logger.log(LogLevel::info, (std::ostringstream()
                                     << "on_recv_datagram cbNotifyQueue size"
                                     << cbNotifyQueue.size()).str());
     }
 
-    if (too_many_in_queue || stream_cnx->rx_data->size() < 2 || stream_cnx->in_data_cb_skip_count > 30) {
+    if (stream_cnx->rx_data->size() < 2 || stream_cnx->in_data_cb_skip_count > 30) {
         stream_cnx->in_data_cb_skip_count = 0;
```
suhasHere, line 1048: any reason for choosing 30 ?
please add a single line comment to help readability

suhasHere (changes requested):
This is looking neat. Suggested few nits

include/transport/safe_queue.h
TimEvens: Thanks for catching that.  Updated.

include/transport/time_queue.h
```diff
@@ -280,7 +287,7 @@ namespace qtransport {
 
             const tick_type ticks = advance();
 
-            const auto future_bucket_index = (_bucket_index + (_duration / _interval) - 1) % _total_buckets;
+            const auto future_bucket_index = (_bucket_index + (_total_buckets - 1) % _total_buckets;
```
RichLogan, line 290: Missing a closing parenthesis here

include/transport/time_queue.h
GhostofCookie: +1

Maybe even include it in the name of the param: `spike_period_interval_ns` (or whatever the unit is)

include/transport/time_queue.h
GhostofCookie: +1

include/transport/time_queue.h
```diff
@@ -314,13 +322,32 @@ namespace qtransport {
...
+            std::lock_guard<std::mutex> lock(_mutex);
+
+            if (_queue_index < _queue.size()) {
+                _queue_index++;
 
-            internal_push(std::move(value), _bucket_index + 1, ticks + 1);
+            } else if (_queue_index) {
+                _queue.clear();
+                _queue_index = 0;
+            }
```
GhostofCookie, line 345: But front is already doing this? As is pop.

include/transport/time_queue.h
```diff
@@ -377,17 +412,54 @@ namespace qtransport {
...
          *        (sliding window)
          *
          * @returns Current tick value at time of advance
          */
-        inline tick_type advance()
+        tick_type advance()
         {
             _timer->get_ticks(Duration_t(_interval), _timer_ctx);
 
+            if (_queue_index && _queue.size() >= _total_buckets && _queue_index >= _queue.size()) {
```
GhostofCookie, line 458:
```cpp
_queue.size() >= _total_buckets
```
This could and absolutely should be true and okay though? The number of buckets does not correlate with the size of of the queue. For instance, every item in the queue could be in bucket 0, let's say we have 5 buckets and a queue size of 100. That's completely fine, so why check this?

As well, let's we have 2 items in the queue, and 5 buckets,  but the queue_index is 2. We've run out the queue, but the number of elements isn't greater than the buckets, so we don't clear (even though we 100% should, since we've consumed the items already).

include/transport/time_queue.h
```diff
@@ -360,11 +387,19 @@ namespace qtransport {
...
 
-                if (value_index >= bucket.size() || ticks > expiry_tick) {
+                if (value_index >= bucket.size() || (ticks > expiry_tick && !in_spike(ticks))) {
+                    std::cerr << "===> front Object has expired"
+                              << " queue_index: " <<_queue_index << " queue_size: " << _queue.size()
+                              << " value_index: " << value_index
+                              << " bucket_index: " << bucket_index
+                              << " bucket_size: " << bucket.size()
+                              << " tick_delta: " << _timer_ctx.delta
+                              << " " << ticks << " > " << expiry_tick
```
GhostofCookie, line 397: Don't love printing in the container, but I suppose we need some info right now? This could get pretty spammy though.

include/transport/time_queue.h
```diff
@@ -333,7 +360,7 @@ namespace qtransport {
                 auto& [bucket_index, value_index, expiry_tick] = _queue.at(_queue_index++);
                 auto& bucket = _buckets.at(bucket_index);
```
GhostofCookie, line 361: Old code I know, but just thinking these should probably be `const auto&`, not sure why they weren't before. Mea culpa.

include/transport/time_queue.h
```diff
@@ -360,11 +387,19 @@ namespace qtransport {
...
+                              << " bucket_index: " << bucket_index
+                              << " bucket_size: " << bucket.size()
+                              << " tick_delta: " << _timer_ctx.delta
+                              << " " << ticks << " > " << expiry_tick
+                              << std::endl;
                     _queue_index++;
                     continue;
                 }
-
+
```
GhostofCookie, line 402: Why are 
```cpp
_queue.clear();
_queue_index = 0;
```
Missing from here? They should be here... for cleanup. Perhaps those missing lines are what necessitated the `move_front` method that was just added?

include/transport/time_queue.h
TimEvens: fixed.

include/transport/time_queue.h
```diff
@@ -377,17 +412,54 @@ namespace qtransport {
...
 
-      private:
+    private:
+        /**
+         * @brief Checks if currently in allowed TTL spike
+         *
+         * @note This should be called only when the front/pop would expire an
+         *      object due to TTL being expired.
+         *
+         * @details Latency spike
```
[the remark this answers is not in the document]
TimEvens: Updated.

include/transport/time_queue.h
```diff
@@ -401,17 +473,10 @@ namespace qtransport {
...
-            if (_queue_index && _queue_index >= _queue.size()) {
-                _queue.clear();
-                _queue_index = 0;
-            }
-
-            for (int i = 0; i < _timer_ctx.delta; i++) {
-                _buckets[(_bucket_index + i) % _total_buckets].clear();
-            }
-
+            _buckets[_bucket_index].clear();
```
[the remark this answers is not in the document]
TimEvens: We need to allow those to be there even though time has passed to support the spike. Otherwise, we clear them and allowing spikes doesn't do anything because the buckets are cleared. :)  I'll add back the clearing when latency spike is disabled.   Clearing one appears to still be okay in terms of memory, but we need to watch it. Buckets are still cleared when latency is greater than bucket size and when the queue is being processed with no lag.

include/transport/time_queue.h
TimEvens: It's relative to the time_queue constructor params `interval` and `duration`.  The value of these settings, including TTL is actually the count of interval.  The default is 1 and the default duration is ms, so these values, including TTL is (value * 1ms).  In this case, changing the param to ms would be fine, but then it would only be fine as long as you never changed those.

Priority queue, for example, locks in 1ms, so those settings are in ms.   Time-queue is customizable and therefore the values are relative to the constructor params. For now, I'll add some comments around this.

include/transport/transport.h
TimEvens: Will update comments, but units are not fixed where the name can be set to say ms. It's defined by the template variable `duration_t`.

src/transport_picoquic.cpp
TimEvens: Updated.

include/transport/time_queue.h
```diff
@@ -377,17 +412,54 @@ namespace qtransport {
...
+         *
+         * @note This should be called only when the front/pop would expire an
+         *      object due to TTL being expired.
+         *
+         * @details Latency spike
+         *
+         * @retun True if currently in spike duration and the TTL should be allowed/not expired
+         *      False if not in spike duration and the object should be expired.
+         */
+         bool in_spike(tick_type current_tick) {
```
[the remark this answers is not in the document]
TimEvens: I was going to put it there, but it's not a timer function. It's only relative to the time-queue itself.

include/transport/time_queue.h
TimEvens: That was one of the bugs.  Depending on time, front was actually popping a message that it shouldn't have been.  If we use front, what we really want to do is just move the index pointer when we want to remove the front object.

include/transport/time_queue.h
TimEvens: This is a debug only. Problem is time queue doesn't have a logger right now and we are still debugging drops. This should really be removed, but I left it here for now so that we can see it happening.

include/transport/time_queue.h
TimEvens: Updated.

include/transport/time_queue.h
TimEvens: It's the same as it was before, after the while loop.  We only move forward up to the max size of the queue. If at the end, it clears/resets it.  Move will also do it if move results in wrap.

include/transport/time_queue.h
GhostofCookie: I think this code should probably go back where it was as it was, this adds some potentially unnecessary checks (queue may be cleared by a jump on line 429). An I'm still not so sure this check against total buckets is correct

include/transport/time_queue.h
```diff
@@ -352,60 +312,74 @@ namespace qtransport {
...
+                     * TODO: Below log is only added for debugging right now. This should be removed when stable or
+                     *    when we have a logger.
+                     */
+                    std::cerr << "===> front Object has expired"
+                              << " queue_index: " <<_queue_index << " queue_size: " << _queue.size()
+                              << " value_index: " << value_index
+                              << " bucket_index: " << bucket_index
+                              << " bucket_size: " << bucket.size()
+                              << " tick_delta: " << _timer_ctx.delta
+                              << " " << ticks << " > " << expiry_tick
```
GhostofCookie, line 332: Probably don't need this now, since without the spike detection this might spam.

GhostofCookie (approved):

include/transport/time_queue.h
TimEvens: Removed.

PR #60, opened 2023-08-25
Introuced the Cantina Logger

paulej:
This PR introduces the Cantina logger which is now in this repository: https://github.com/Quicr/logger.

The logs that are debug logs use the macro `LOGGER_DEBUG()` so that we can (if desired) compile those log messages out of the builds.  I did not put anything in place to explicitly do that, but it is possible.

Most of the logging uses the streaming operators, while some use the `Log()` function.  It all results in the same underlying function getting called, but the streaming operators avoid the need to use `std::ostringstring`.  There is use of the latter, still, because the composed string is used in logging and throwing an exception.


cmd/echoServer.cc
```diff
@@ -3,19 +3,23 @@
 #include <sstream>
 #include <thread>
 
-#include "cmdLogger.h"
 #include <transport/transport.h>
+#include "cantina/logger.h"
```
GhostofCookie, line 7: Shouldn't this be using <> instead of ""?

src/transport_picoquic.cpp
```diff
@@ -215,9 +214,7 @@ pq_event_cb(picoquic_cnx_t* cnx,
         }
 
         default:
-            log_msg.str("");
-            log_msg << "Got event " << fin_or_event;
-            transport->logger.log(LogLevel::debug, log_msg.str());
+            LOGGER_DEBUG(transport->logger, "Got event " << fin_or_event);
```
GhostofCookie, line 217: Should all the calls to `transport->logger->info` to `LOGGER_INFO()`?

paulej: It depends on whether we want the ability to compile those out.  I just changed the debug logs since those are quite likely something we want to compile out.  The others we are more likely to want to make a run-time option.

cmd/echoServer.cc
paulej: No. The others should be "transport/transport.h".  The <> is for system include files, whereas "" is for non-system include files.

paulej: Changed.

cmd/client.cc
```diff
@@ -39,16 +39,12 @@ struct Delegate : public ITransport::TransportDelegate
     void on_connection_status(const TransportContextId& context_id, const TransportStatus status)
     {
         tcid = context_id;
-        std::stringstream s_log;
-        s_log << "Connection state change context: " << context_id << ", " << int(status);
-        logger.log(LogLevel::info, s_log.str());
+        logger->info << "Connection state change context: " << context_id << ", " << int(status) << std::flush;
```
TimEvens, line 42: Is `std::flush` always needed?

cmd/client.cc
```diff
@@ -103,23 +97,21 @@ main()
 
     auto client = ITransport::make_client_transport(server, tconfig, d, logger);
 
-    logger.log(LogLevel::info, "client use_count: " + std::to_string(client.use_count()));
+    logger->info << "client use_count: " << std::to_string(client.use_count()) << std::flush;
 
     d.setClientTransport(client);
-    logger.log(LogLevel::info, "after set client transport client use_count: " + std::to_string(client.use_count()));
+    logger->info << "after set client transport client use_count: " << std::to_string(client.use_count()) << std::flush;
```
TimEvens, line 103: Do we still need `to_string()`?

src/transport_picoquic.cpp
```diff
@@ -246,7 +244,7 @@ pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* ca
         switch (cb_mode) {
             case picoquic_packet_loop_ready: {
                 log_msg << "packet_loop_ready, waiting for packets";
-                transport->logger.log(LogLevel::info, log_msg.str());
+                transport->logger->info << log_msg.str() << std::flush;
```
TimEvens, line 247: Can we drop `log_msg` and append to logger info with flush at the end?

src/transport_picoquic.cpp
```diff
@@ -271,7 +269,7 @@ pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* ca
 
             case picoquic_packet_loop_port_update:
                 log_msg << "packet_loop_port_update";
-                transport->logger.log(LogLevel::debug, log_msg.str());
+                LOGGER_DEBUG(transport->logger, log_msg.str());
```
TimEvens, line 272: the macros are nice but maybe we use them consistently... otherwise maybe not use them.

src/transport_picoquic.cpp
```diff
@@ -466,8 +459,9 @@ PicoQuicTransport::PicoQuicTransport(const TransportRemote& server,
                                      const TransportConfig& tcfg,
                                      TransportDelegate& delegate,
                                      bool _is_server_mode,
-                                     LogHandler& logger)
-  : logger(logger)
+                                     const cantina::LoggerPointer& logger)
+  : logger(std::make_shared<cantina::Logger>("QUIC", logger))
```
TimEvens, line 463: So each class will call `make_shared()` instead of assignment of logger to plus ref count?

src/transport_picoquic.cpp
```diff
@@ -572,15 +594,47 @@ PicoQuicTransport::start()
     uint64_t current_time = picoquic_current_time();
 
     if (debug) {
-        debug_set_stream(stdout); // Enable picoquic debug
+        if (!opened_logging_fds)
+        {
+            // Open pipe for reading and writing
+            if (pipe(logging_fds) == 0) {
```
TimEvens, line 600: Need to watch to ensure we don't introduce a blocking operation here.

src/transport_picoquic.cpp
```diff
@@ -623,6 +677,58 @@ PicoQuicTransport::start()
...
+
+        // Wait until something is ready to read
+        if (select(logging_fds[0] + 1,
+                    &pipe_fds,
+                    nullptr,
+                    nullptr,
+                    nullptr) == -1) {
+            pico_logger->error << "Select for logging failed" << std::flush;
+            // Set picoquic debug to stdout
+            debug_set_stream(stdout);
```
TimEvens, line 702: Normally it's unsafe to call any picoquic functions outside of the picoquic loop thread from another thread. It's probably okay for now... but.... any reason why this isn't using the pq runner queue?

src/transport_picoquic.h
```diff
@@ -165,13 +165,19 @@ class PicoQuicTransport : public ITransport
     /*
    * Internal Public Variables
      */
-    LogHandler& logger;
+    cantina::LoggerPointer logger;
+    int logging_fds[2];
```
TimEvens, line 169: Maybe add a comment describing where/why this is used.

cmd/client.cc
paulej: If using the streaming operator, yes. It flushes the buffer. It's like having `std::endl`.  But `std::endl` is used to insert new lines, so was not used.  If using the functions called (`logger.Log()`) or the macros like `LOGGER_DEBUG`, then std::flush is performed automatically.

cmd/client.cc
paulej: No. I'll remove that.

src/transport_picoquic.cpp
paulej: We could, but log_msg is populated with a prefix string above this line of code that is used in multiple case statements. Thus, it made sense to leave it as-is.

src/transport_picoquic.cpp
paulej: Most log messages we will likely want to keep in release code and enable with a run-time config.  However, debug logs are likely ones we want to compile-out.  That was the reason I used LOGGER_DEBUG. In general, I'm not a fan of the macro form, but I do appreciate the ability to compile out those lines.

src/transport_picoquic.cpp
paulej: Our choice.  We can make a copy of the logger pointer or create a new pointer with a parent-child relationship (as is the case here). The benefit of the parent-child relationship is that each module will have its own module name (here `QUIC`) that appears in the logs.  Each of the parent logging object prepend their module names, too, making it easy to identify exactly what object emitted the log message.

src/transport_picoquic.cpp
paulej: There ought not be one.  There is a separate thread reading the logs written into the pipe.

src/transport_picoquic.cpp
paulej: This is in the same thread where `debug_set_stream()` was already called. I didn't introduce anything new here.

src/transport_picoquic.h
paulej: I'm most definitely not opposed to comments, but there aren't others here. Personally, I like a simple comment on every single variable, but most have nothing. I didn't want to create inconsistencies.

GhostofCookie (approved):
Nothing really scares me about this code (except maybe the pipe/select stuff), so I think this is good to go!

TimEvens (approved):
LGTM... better to get this in now before the other changes.

paulej:
@TimEvens I'd like to, but laps will not build against the current transport code.  I'm not sure who is making changes where, but I think that needs to be building against the current transport code before we merge this.

PR #62, opened 2023-08-30
Refactor queues

GhostofCookie:
Noticed some stuff in time_queue that could be reduced and it kind of ran away from me. Also matched up safeQueue naming with priority_queue and time_queue (safe_queue), and modernized some code in priority_queue.


TimEvens (approved):
LGTM. We'll just need to update upstream repos, such as laps, to support the name change.

PR #63, opened 2023-09-01
initial PR to toggle quic support and add esp threading

suhasHere:
This PR brings in minimal support to 
   - toggle building transport with QUIC as optional
   -  add few esp api to get thread stack configuration for hactar messaging

The latter of the 2 will be revisited once we have basic integration working

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,9 +9,11 @@ project(quicr-transport
 if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     option(QTRANSPORT_BUILD_TESTS "Build tests for quicr transport" ON)
 else()
-    option(QTRANSPORT_BUILD_TESTS "Build tests for quicr trasnsport" OFF)
+    option(QTRANSPORT_BUILD_TESTS "Build tests for quicr transport" ON)
 endif()
 
+option(QUIC_TRANSPORT "Enable QUIC Transport (Default ON)" OFF)
+option(PLATFORM_ESP_IDF "Enabble suppport for esp-idf (Default OFF" OFF)
 
 find_package(Threads REQUIRED)
 find_package(PkgConfig REQUIRED)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -9,30 +9,49 @@ project(quicr-transport
 if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     option(QTRANSPORT_BUILD_TESTS "Build tests for quicr transport" ON)
 else()
-    option(QTRANSPORT_BUILD_TESTS "Build tests for quicr trasnsport" OFF)
+    option(QTRANSPORT_BUILD_TESTS "Build tests for quicr transport" OFF)
 endif()
 
+option(QUIC_TRANSPORT "Enable QUIC Transport (Default ON)" ON)
```
TimEvens, line 15: Considering selection of UDP vs QUIC vs H3 vs maybe TCP in the future is at execution time, disabling QUIC or any other transport protocol option would invalidate the execution time config and use-cases to support that.  

What was the reason to not compile in the other protocols? Is it that the dependencies with QUIC are too much causing pain for testing? If so, maybe we wrap this into TEST knobs.

src/transport_udp.cpp
```diff
@@ -524,8 +553,27 @@ UDPTransport::connect_client()
   // Notify caller that the connection is now ready
   delegate.on_connection_status(last_context_id, TransportStatus::Ready);
 
+  #if defined(PLATFORM_ESP)
+  auto cfg = create_config("FDReader", 1, 12 * 1024, 5);
```
TimEvens, line 557: Maybe add transport config options?

CMakeLists.txt
```diff
@@ -9,30 +9,49 @@ project(quicr-transport
...
+add_library(quicr-transport ${sources})
+
+if (QUIC_TRANSPORT)
+  target_link_libraries(quicr-transport
+          PUBLIC
+          picoquic-core)
+   add_compile_definitions(${LIB_NAME} ENABLE_QUIC)       
+endif()          
+
+if (PLATFORM_ESP_IDF)
```
TimEvens, line 51: ```/Users/timevens/Documents/workspace/ws-media10x/transport/src/transport_udp.cpp:20:10: fatal error: 'lwip/netdb.h' file not found```

Looks like we need to include submodule with the new headers used.

TimEvens (commented):
ESP-IDF doesn't currently build for me because of missing headers. 

Curious as to why to use ESP-IDF for just setting thread priority vs `native_handle()` and `pthread_setschedparam()`? Is the thought we'll use more of esp-idf?

CMakeLists.txt
suhasHere: this is a temporary fix to get  esp builds going. PicoTLS has been updated to support mbed tls then we don't need this anymore.

CMakeLists.txt
suhasHere: this option is enabled when building with esp framework and not as submodules. So we never enable this option for anything outside that setup and hence would be ok.

src/transport_udp.cpp
suhasHere: good idea. but these are test parameters to verify the functionality. Once we have the builds integrated and can test the options, happy to move it as config option

CMakeLists.txt
TimEvens: What's the make file or other syntax to build this?  Right now it doesn't build for me.

suhasHere: you can't build it unless you are building within esp framework. As long as transport builds with no PLATFORM_ESP_IDF,  we should be fine to merge this PR

TimEvens (approved):
LGTM, nice to try it out.

PR #64, opened 2023-09-05
Changes to support picoquic logging callbacks

paulej:
This will use callbacks to receive logging from picoquic.  **This does not currently build**.

This code is waiting on this PR: https://github.com/private-octopus/picoquic/pull/1544

Once that PR is merged and the submodules in the quicr/transport code is updated to receive that PR, this code should work.


TimEvens (approved):
LGTM, but we need to wait on this till we sync/merge the picoquic changes.

PR #66, opened 2023-09-05
Updated cantina Logger

paulej:
This updates the cantina logger so that one may configure turning off syslog support.


TimEvens (approved):
LGTM

PR #71, opened 2023-09-14
Latency spikes

TimEvens:
Adds transport config and picoquic config option to set CWIN. 

This also adds a log message for **CB delta** for us to track for now delayed callbacks.  This log message only logs if the delta from last CB to current CB (per stream/datagram) is greater than 40ms.  40ms delay is pretty large, hopefully we don't see this that often.


suhasHere (approved):

src/transport_picoquic.cpp
```diff
@@ -783,6 +786,7 @@ PicoQuicTransport::createClient()
     struct sockaddr_storage server_address;
     char const* sni = "cisco.webex.com";
     int ret;
+    std::ostringstream log_msg;
```
GhostofCookie, line 789: What's this for?

TimEvens: Removed.

GhostofCookie (approved):

GhostofCookie (approved):

src/transport_picoquic.cpp
```diff
@@ -1161,7 +1173,7 @@ void PicoQuicTransport::on_recv_stream_bytes(StreamContext* stream_cnx, uint8_t*
     uint8_t *bytes_p = bytes;
 
     if (stream_cnx == NULL || length == 0) {
-        logger->Log(cantina::LogLevel::Warning, "on_recv_stream_bytes has null context");
+        logger->Log(cantina::LogLevel::Debug, "on_recv_stream_bytes has null context");
```
paulej, line 1176: Since you want this to be debug, perhaps use the `LOGGER_DEBUG` macro so this line can be compiled out? It won't matter much if this line is rarely hit, but if seen often it will make a difference in production code.

TimEvens: Thanks. I think there are a few places we need to update the logging. Might be best to do that as a logging PR.

PR #73, opened 2023-10-23
Update picoquic with minor updates to transport

TimEvens:
* Fix to support IPv6 (MDNS will not work with IPv6)
* Tune based on feedback from Christian
* Disables logging of metrics considering it's too much data to print with debug prints and we will be moving this to influxDb/exports.

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -9,7 +9,6 @@
 #include <ctime>
 #include <iostream>
 #include <netdb.h>
-#include <sstream>
 #include <thread>
 #include <unistd.h>
 
@@ -106,15 +105,10 @@ pq_event_cb(picoquic_cnx_t* cnx,
             break;
 
         case picoquic_callback_datagram_spurious:
-            // TODO: Add metrics for spurious datagrams
-            // delayed ack
-            // std::cout << "spurious DGRAM length: " << length << std::endl;
             transport->metrics.dgram_spurious++;
             break;
 
         case picoquic_callback_datagram_lost:
-            // TODO: Add metrics for lost datagrams
-            // std::cout << "Lost DGRAM length " << length << std::endl;
             transport->metrics.dgram_lost++;
             break;
 
@@ -175,7 +169,6 @@ pq_event_cb(picoquic_cnx_t* cnx,
             transport->logger->info << std::flush;
 
             picoquic_set_callback(cnx, NULL, NULL);
-            picoquic_close(cnx, 0);
             transport->deleteStreamContext(reinterpret_cast<uint64_t>(cnx), stream_id);
 
             if (not transport->_is_server_mode) {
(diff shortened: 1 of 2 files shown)


src/transport_picoquic.cpp
```diff
@@ -370,7 +367,7 @@ PicoQuicTransport::deleteStreamContext(const TransportContextId& context_id, con
         StreamContext* s_cnx = &active_streams[context_id][stream_id];
 
         on_connection_status(s_cnx, TransportStatus::Disconnected);
-        picoquic_close_immediate(s_cnx->cnx);
+        picoquic_close(s_cnx->cnx, 0);
```
TimEvens, line 370: Christian confirmed that close is what we should be using.

paulej (approved):

src/transport_picoquic.cpp
```diff
@@ -338,7 +335,7 @@ pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void* ca
                         close_cnx = picoquic_get_next_cnx(close_cnx);
                     }
 
-                    return PICOQUIC_NO_ERROR_TERMINATE_PACKET_LOOP;
+                    return 0; //PICOQUIC_NO_ERROR_TERMINATE_PACKET_LOOP; picoquic_close() will stop the loop
```
TimEvens, line 338: Do not close directly here since we need the loop to finish.

src/transport_picoquic.cpp
```diff
@@ -658,8 +655,9 @@ PicoQuicTransport::start()
     picoquic_init_transport_parameters(&local_tp_options, 1);
     local_tp_options.max_datagram_frame_size = 1280;
     //  local_tp_options.max_packet_size = 1450;
-    //  local_tp_options.max_ack_delay = 3000000;
-    //  local_tp_options.min_ack_delay = 500000;
+    local_tp_options.idle_timeout = 10000;
+    local_tp_options.max_ack_delay = 100000;
```
TimEvens, line 659: Adjustments based on feedback from Christian.

src/transport_picoquic.cpp
```diff
@@ -847,9 +845,8 @@ PicoQuicTransport::client(const TransportContextId tcid)
             return;
         }
 
-        ret = picoquic_packet_loop(quic_ctx, 0, AF_INET, 0, 2000000, 0, pq_loop_cb, this);
+        ret = picoquic_packet_loop(quic_ctx, 0, PF_UNSPEC, 0, 2000000, 0, pq_loop_cb, this);
```
TimEvens, line 848: Support IPv6 and IPv4!!!

src/transport_picoquic.cpp
```diff
@@ -1179,6 +1176,12 @@ void PicoQuicTransport::on_recv_stream_bytes(StreamContext* stream_cnx, uint8_t*
 
     bool object_complete = false;
 
+    if (stream_cnx->stream_rx_object != nullptr && stream_cnx->stream_rx_object_size == 0) {
```
TimEvens, line 1179: Debug for now to report this condition that really shouldn't happen.

src/transport_picoquic.cpp
paulej: For the server (accepting connections), it would need to listen for packets on both IPv4 and IPv6 (thus, using two file descriptors). Does it do that?

PR #74, opened 2023-10-27
Fix stream priority setting and update callback delays to support streams

TimEvens:


src/transport_picoquic.cpp
```diff
@@ -559,10 +574,14 @@ PicoQuicTransport::createStream(const TransportContextId& context_id,
     next_stream_id = ::make_stream_id(next_stream_id + 4, _is_server_mode, _is_unidirectional);
 
     PicoQuicTransport::StreamContext* stream_cnx = createStreamContext(cnx_stream_iter->second.cnx, next_stream_id);
+    stream_cnx->priority = priority;
 
+    /*
+     * Low order bit set indicates FIFO handling of samem priorities, unset is round-robin
```
GhostofCookie, line 580: samem?

src/transport_picoquic.cpp
```diff
@@ -786,13 +805,32 @@ PicoQuicTransport::close([[maybe_unused]] const TransportContextId& context_id)
...
 
 void
-PicoQuicTransport::send_next_datagram(StreamContext* stream_cnx, uint8_t* bytes_ctx, size_t max_len)
-{
-    static auto last_time = std::chrono::steady_clock::now();
+PicoQuicTransport::check_callback_delta(StreamContext* stream_cnx, bool tx) {
     auto now_time = std::chrono::steady_clock::now();
-    auto delta_ms = std::chrono::duration_cast<std::chrono::milliseconds>(now_time - last_time).count();
 
+    if (tx) {
```
GhostofCookie, line 811: Maybe 
```cpp
if (!tx) return;
```
?

GhostofCookie (approved):

src/transport_picoquic.cpp
TimEvens: updated

src/transport_picoquic.cpp
TimEvens: I can put that in, but the thought is that we'll have an else{} later.  We can add that later.

PR #76, opened 2023-11-08
Fixes possible memory leak and adds retransmit debug logging

TimEvens:
* Fix cleaning up streams on transport destructor and stream close Resolves possible memory leak with tx/rx stream objects not being freed.
* Add debug/tracking logging of retransmits per connection
* Move out of stream context peer/remote IP and port to connection context
* Update related code to use connection context
* Make statics member variables in pq_loop_cb

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -169,7 +169,9 @@ pq_event_cb(picoquic_cnx_t* cnx,
                                     << stream_id;
 
             if (stream_cnx != NULL) {
-                transport->logger->info << " " << stream_cnx->peer_addr_text;
+                auto conn_ctx = transport->getConnContext(reinterpret_cast<uint64_t>(cnx));
+
+                transport->logger->info << " " << conn_ctx.peer_addr_text;
             }
 
             transport->logger->info << std::flush;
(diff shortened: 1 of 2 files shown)


TimEvens:
In other PR, I'll rename `context_id` and related to `conn_id`.  For example, `TransportContextId` is really `TransportConnectionId`.  It was context from way back when we first started this with the stub code.  There's also some other naming I would like to update, such as the use of `cnx` vs `ctx`.  Anyhow, that's for a different PR.

src/transport_picoquic.cpp
```diff
@@ -169,7 +169,9 @@ pq_event_cb(picoquic_cnx_t* cnx,
                                     << stream_id;
 
             if (stream_cnx != NULL) {
-                transport->logger->info << " " << stream_cnx->peer_addr_text;
+                auto conn_ctx = transport->getConnContext(reinterpret_cast<uint64_t>(cnx));
+
+                transport->logger->info << " " << conn_ctx.peer_addr_text;
```
paulej, line 174: Just a desire to print the peer IP address?  If this connection was already established, I think code below would have retrieved this address. Is this for cases where the connection is not fully established?

paulej (approved):

GhostofCookie (approved):

src/transport_picoquic.cpp
TimEvens: It's just for info because a close may end up being way down in the logs. It's nice to have the IP along with it. Considering this is just for close, it should be fine in terms of perf.  Although, if we start to add IP to other logs, we might want to put back the IP/port string in the stream context to support faster logging.  We might not need to add the IP log messages in the future, except for initial connect considering it doesn't add much value when multiple clients are behind NAT/PAT. It'll then be down to port, which isn't much better than looking at connection ID (context_id).

PR #77, opened 2023-11-09
Fix segfaults caused by bad memory access on close

TimEvens:


src/transport_picoquic.cpp
```diff
@@ -374,13 +376,14 @@ PicoQuicTransport::deleteStreamContext(const TransportContextId& context_id, con
         StreamContext* s_cnx = &active_streams[context_id][stream_id];
 
         on_connection_status(s_cnx, TransportStatus::Disconnected);
-        picoquic_runner_queue.push([=, cnx = s_cnx->cnx]() {
-            picoquic_close(cnx, 0);
-        });
+
+        picoquic_close(s_cnx->cnx, 0);
```
TimEvens, line 380: Still not 100% thread safe, but this should only be called when stream zero is being closed by the pq event loop.  Deferring this in pq_runner caused bad access to the CNX pointer.

src/transport_picoquic.cpp
```diff
@@ -374,13 +376,14 @@ PicoQuicTransport::deleteStreamContext(const TransportContextId& context_id, con
...
 
         // Remove all streams if closing the root/datagram stream (closed connection)
         active_streams.clear();
 
+        conn_context.erase(context_id);
```
TimEvens, line 385: The connection context needed to be removed when stream (datagram) zero is closed. Not removing it resulted in bad access to cnx in the context.

src/transport_picoquic.cpp
```diff
@@ -1150,13 +1153,7 @@ void PicoQuicTransport::on_recv_stream_bytes(StreamContext* stream_cnx, uint8_t*
...
     bool object_complete = false;
 
-    if (stream_cnx->stream_rx_object != nullptr && stream_cnx->stream_rx_object_size == 0) {
-        logger->warning <<  "on_recv_stream_bytes has object size zero and non-null rx_object sid: "
-                        << std::to_string(stream_cnx->stream_id) << std::flush;
-
-    }
-
-    if (stream_cnx->stream_rx_object == nullptr || stream_cnx->stream_rx_object_size == 0) {
+    if (stream_cnx->stream_rx_object == nullptr) {
```
TimEvens, line 1156: Only nullptr should be allowed.  Delete later in this block would error out if null but object size is zero.

paulej (approved):

paulej (approved):

GhostofCookie (approved):

PR #78, opened 2023-11-16
Offsite changes

TimEvens:
* Updates picoquic to latest master
* Fixes segfaults
* Adds shadow RTT
* Minor perf updates to reduce CPU
* Reduces duplicate code with rx/tx object buffers
* Adds support for RX streams to support callbacks less than header size
*

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -373,6 +373,11 @@ PicoQuicTransport::deleteStreamContext(const TransportContextId& context_id, con
 
         on_connection_status(s_cnx, TransportStatus::Disconnected);
 
+        // Remove pointer references in picoquic for active streams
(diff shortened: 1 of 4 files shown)


src/transport_picoquic.cpp
```diff
@@ -601,7 +602,7 @@ PicoQuicTransport::createStream(const TransportContextId& context_id,
 
     const auto& iter = active_streams.find(context_id);
     if (iter == active_streams.end()) {
-        throw std::invalid_argument("Invalid context id, cannot create stream. context_id = " + std::to_string(context_id));
+        logger->info << "Invalid context id, cannot create stream. context_id = " << context_id << std::flush;
```
GhostofCookie, line 605: Return?

GhostofCookie (approved):

PR #107, opened 2023-12-13
Resolves #106

### LINKED ISSUE #106: Invalid hostname does not return error
When client provides an invalid hostname, the transport does not return an error. Instead it attempts to connect to something resulting in a long delay for idle timeout.

TimEvens:


src/transport_picoquic.cpp
```diff
@@ -491,8 +493,9 @@ PicoQuicTransport::start()
         logger->info << "Connecting to server " << serverInfo.host_or_ip << ':' << serverInfo.port
                      << std::flush;
 
-        cid = createClient();
-        picoQuicThread = std::thread(&PicoQuicTransport::client, this, cid);
+        if ((cid = createClient())) {
```
paulej, line 496: I was curious about the return type here. Given the return type is a pointer that is typecast, a safer definition for `TransportContextId` would be `uintptr_t`. That is guaranteed to be an integer type of proper size to hold a pointer regardless of the processor's word size.

paulej (approved):

src/transport_picoquic.cpp
TimEvens: We are returning a copy of the `uni64_t`, which will always be zero or some positive `uin64_t`. **TransportConnId** is a `uint64_t` that is the 64bit pointer memory address value. It is **not used** to def back to a pointer.  It's only used as the numeric value for the connection Id. `reinterpret_cast<TransportConnId>(pq_cnx) is only converting it to a persistent number that can be used as an id.

Issue #110, opened 2024-01-08
Add support for IPv6 in UDP transport


TimEvens:
Add/fix support for IPv6 in UDP transport.

PR #111, opened 2024-01-12
Refactor UDP transport

TimEvens:
Significant refactor to add the following new features:

* Connection establishment - helps with port scanning/dos mitigation
* Connection disconnect - Client disconnect is now signaled to close properly
* Connection idle timeout - If the client or server disappears (e.g., crash), idle timeout
  will clean up the connection. 
* Shipping/pacing transmission to configurable rate per microsecond (applied at the millisecond level). 
* Added report feedback with logging and initial calculation of bandwidth and loss. This will be tuned and adjusted
  based on testing. We can use the feedback report to adjust the shaping/pacing.

### DIFF
--- a/include/transport/priority_queue.h
+++ b/include/transport/priority_queue.h
@@ -43,7 +43,6 @@ namespace qtransport {
 
       public:
         ~priority_queue() {
-            std::lock_guard<std::mutex> _(_mutex);
         }
 
         /**
--- a/include/transport/priority_queue.h
+++ b/include/transport/priority_queue.h
@@ -42,6 +42,10 @@ namespace qtransport {
         };
 
       public:
+        ~priority_queue() {
+            std::lock_guard<std::mutex> _(_mutex);
+        }
+
         /**
          * Construct a priority queue
          * @param tick_service Shared pointer to tick_service service
@@ -154,13 +158,13 @@ namespace qtransport {
         /**
          * @brief Clear queue
          */
-        void clear() {
+         void clear()
+        {
             std::lock_guard<std::mutex> _(_mutex);
 
-            for (auto& tqueue: _queue) {
-                if (tqueue && !tqueue->empty()) {
+            for (auto& tqueue : _queue) {
+                if (tqueue && !tqueue->empty())
                     tqueue->clear();
-                }
             }
         }
 
--- a/src/transport_udp.cpp
+++ b/src/transport_udp.cpp
@@ -26,7 +26,7 @@ UDPTransport::~UDPTransport()
   stop = true;
 
   // Clear threads from the queue
-  fd_write_queue.stop_waiting();
+  fd_write_queue.reset();
 
   // Close socket fd
   if (fd >= 0)
@@ -37,6 +37,8 @@ UDPTransport::~UDPTransport()
     if (thread.joinable())
         thread.join();
   }
+
+  _tick_service.reset();
 }
 
 UDPTransport::UDPTransport(const TransportRemote& server,
@@ -50,6 +52,9 @@ UDPTransport::UDPTransport(const TransportRemote& server,
   , serverInfo(server)
   , delegate(delegate)
 {
+  _tick_service = std::make_shared<threaded_tick_service>();
+  fd_write_queue = std::make_unique<priority_queue<ConnData>>(
+          1000, 1, _tick_service, 1000);
 }
 
 TransportStatus
@@ -70,13 +75,13 @@ UDPTransport::createDataContext(
   [[maybe_unused]] bool bidir)
 {
 
-  if (remote_contexts.count(conn_id) == 0) {
-    logger->error << "Invalid context id: " << conn_id << std::flush;
+  if (connections.count(conn_id) == 0) {
+    logger->error << "Invalid connection id: " << conn_id << std::flush;
     return 0; // Error
   }
 
-  auto addr = remote_contexts[conn_id];
-  return remote_addrs[addr.key].sid;
+  auto conn = connections[conn_id];
+  return remote_addrs[conn.addr.key].sid;
 }
 
 TransportConnId
@@ -102,14 +107,14 @@ UDPTransport::deleteDataContext(const TransportConnId& context_id, DataContextId
 }
 
 bool
-UDPTransport::getPeerAddrInfo(const TransportConnId& context_id,
+UDPTransport::getPeerAddrInfo(const TransportConnId& conn_id,
                               sockaddr_storage* addr)
 {
   // Locate the given transport context
-  auto it = remote_contexts.find(context_id);
+  auto it = connections.find(conn_id);
 
   // If not found, return false
-  if (it == remote_contexts.end()) return false;
+  if (it == connections.end()) return false;
 
   // Copy the address information
   std::memcpy(addr, &it->second.addr, sizeof(it->second.addr));
@@ -118,17 +123,17 @@ UDPTransport::getPeerAddrInfo(const TransportConnId& context_id,
 }
 
 void
-UDPTransport::close(const TransportConnId& context_id)
+UDPTransport::close(const TransportConnId& conn_id)
 {
 
   if (isServerMode) {
     addrKey ak;
 
-    addr_to_key(remote_contexts[context_id].addr, ak);
+    addr_to_key(connections[conn_id].addr.addr, ak);
 
     remote_addrs.erase(ak);
-    remote_contexts.erase(context_id);
-    dequeue_data_map.erase(context_id);
+    connections.erase(conn_id);
+    dequeue_data_map.erase(conn_id);
   }
 }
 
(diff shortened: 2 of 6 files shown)


src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
+
+        if (data[0] != UdpProtocol::PROTOCOL_VERSION) {
+            // TODO: Add metrics to track discards on invalid received message
+            continue;
+        }
+
+        const auto current_tick = _tick_service->get_ticks(std::chrono::milliseconds(1));
+
+        remote_addr.id = create_addr_id(remote_addr.addr);
+        const auto a_conn_it = addr_conn_contexts.find(remote_addr.id);
```
paulej, line 570: Isn't a lock needed to protect access to `conn_contexts`?

src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
+    if (bytes.empty()) {
+        return TransportError::None;
     }
 
-    std::vector<uint8_t> buffer (data, data + rLen);
 
-    connData cd;
-    cd.data = buffer;
-    cd.streamId = 0;
+    const auto conn_it = conn_contexts.find(conn_id);
```
paulej, line 769: Is a lock needed to protect this map?

src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
+}
 
-        cd.contextId = last_context_id;
-        cd.streamId = last_stream_id;
+std::optional<std::vector<uint8_t>> UDPTransport::dequeue(const TransportConnId &conn_id,
+                                                          const DataContextId &data_ctx_id) {
 
-        // Create dequeue
-        dequeue_data_map[last_context_id][last_stream_id].set_limit(1000);
+    const auto conn_it = conn_contexts.find(conn_id);
```
paulej, line 796: Is a lock needed to protect this map?

src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
-      delegate.on_recv_notify(cd.contextId, cd.streamId);
+    fd = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+    if (fd == -1) {
+        throw std::runtime_error("socket() failed");
     }
-  }
 
-  logger->Log("Done transport reader thread");
-}
+    size_t snd_rcv_max = 64000;   // TODO: Add config for value
```
paulej, line 827: I saw a couple of places where 64000 was explicitly referenced. I'd suggest making this `constexpr size_t Max_Packet_Size = 65535` or something until we sort out the dynamic configuration.

src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
+    serverAddr.id = create_addr_id(serverAddr.addr);
 
-  if (dq.size() <= 0) {
-    return std::nullopt;
-  }
+    ++last_conn_id;
 
-  return dq.pop().value().data;
-}
+    const auto& [conn_it, _] = conn_contexts.emplace(last_conn_id,
```
paulej, line 910: Is a lock needed to protect this map?

src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
+    int err =
+            setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, (const char *) &one, sizeof(one));
+    if (err != 0) {
+        s_log << "connect_server: setsockopt error: " << strerror(errno);
+        logger->Log(cantina::LogLevel::Critical, s_log.str());
+        throw std::runtime_error(s_log.str());
+    }
+
+    // TODO: Add config for this value
+    size_t snd_rcv_max = 2000000;
```
paulej, line 955: Above, I saw a value about half this. Another good candidate to make a global constant for the moment.

src/transport_udp.cpp
```diff
@@ -242,361 +520,480 @@ UDPTransport::fd_writer()
...
+            setsockopt(fd, SOL_SOCKET, SO_RCVBUF, &snd_rcv_max, sizeof(snd_rcv_max));
+    if (err != 0) {
+        s_log << "client_connect: Unable to set receive buffer size: "
+              << strerror(errno);
+        logger->Log(cantina::LogLevel::Critical, s_log.str());
+        throw std::runtime_error(s_log.str());
+    }
+
+    err =
+            setsockopt(fd, SOL_SOCKET, SO_RCVTIMEO, &rcv_timeout, sizeof(rcv_timeout));
```
paulej, line 977: Is this needed on a UDP socket?

src/transport_udp_protocol.h
```diff
@@ -0,0 +1,104 @@
...
+         *
+         * @details UDP Protocol starts off with a connect message. Messages will be discarded by the remote
+         *      until the new connection sends a connect message.
+         */
+        struct ConnectMsg : CommonHeader {
+            ConnectMsg() { type = ProtocolType::CONNECT; }
+
+            uint16_t idle_timeout { 120 };            /// Idle timeout in seconds. Must not be zero
+
+        } __attribute__((__packed__, aligned(1)));
```
paulej, line 44: Ah, this explains what I was asking about before. This does ensure the size is precisely the size of the object. Why are these not serialized using the message buffer stuff used elsewhere in libquicr? Perhaps because this is one level down from that, but you know this will not be portable.  It would be trivial to create a miniature message serializer for these, since they're all trivial.

paulej (approved):
I made a number of comments, though you're free to accept or ignore them. :)

include/transport/priority_queue.h
```diff
@@ -42,6 +42,10 @@ namespace qtransport {
         };
 
       public:
+        ~priority_queue() {
+            std::lock_guard<std::mutex> _(_mutex);
```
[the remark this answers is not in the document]
TimEvens: Destruct starts from another thread, which caused a race condition on a pop/push that was happening at the same time.

src/transport_udp.cpp
```diff
@@ -1,237 +1,515 @@
...
+    logger->Log("Starting transport writer thread");
+
+    bool sent_data = false;
+    int all_empty_count = 0;
+
+    while (not stop) {
+        sent_data = false;
+
+
+        std::unique_lock<std::mutex> lock(_connections_mutex);
```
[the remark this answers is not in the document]
TimEvens: It's unlocked because close() locks as well. It needs to because close can be called by another thread as well.   This removes deadlock in that case, which I was hitting. :) 

I'll move the unique lock out of the while loop and then lock on each iteration. This is basically protecting connections map over all threads. 

Unlock is before sent data and I added lock after the select. Thanks for the feedback on that.

src/transport_udp.cpp
TimEvens: Yes, thanks. I've added that.

src/transport_udp.cpp
TimEvens: Yeah, it would be good to add a lock here too.  Updated.

src/transport_udp.cpp
TimEvens: Yep, added.

src/transport_udp.cpp
TimEvens: Good suggestion, updated.

src/transport_udp.cpp
TimEvens: Not on this one because it's creating the client.  Although, we can add a lock for the crazy use-case where the app calls start and close at the same time from two different threads.

src/transport_udp.cpp
TimEvens: Max packet size is pretty locked in, but it would make sense to update all the others that should be config variables in another PR. I would like to do that very soon after this.

src/transport_udp.cpp
TimEvens: It's used today so that we don't block on reading/waiting for data.

src/transport_udp_protocol.h
TimEvens: We can update it if it's really a problem, but:

1. It's portable with clang and gcc of versions that support c++20, which we require.
2. `#pragma pack(1)` is portable with clang, gcc, and vs, so we could use that as well, but that's header wide

TimEvens:
@paulej , thanks so much for the review. I updated all the comments and code based on feedback.  The one change not updated was the packing since it is portable to clang and gcc, which is what we require right now.

include/transport/priority_queue.h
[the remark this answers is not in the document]
paulej: I assume this object is held with a shared pointer. Whatever threads are calling into it should have a valid pointer (i.e., the object's destructor has not been called). Do you know what threads? While this might appear to fix it, it concerns me that we're not first terminating those threads before attempting to destroy the priority queue object.

src/transport_udp.cpp
paulej: This should put the socket in non-blocking mode:
```c
auto flags = fcntl(new_socket, F_GETFL);
fcntl(new_socket, F_SETFL, flags | O_NONBLOCK);
```

Each of those calls can return an error that needs to be checked.

TimEvens: Right now we specifically do not want it in blocking mode. That's the reason we have the timeout and the dedicated threads for reader/writer. 

We can change it to non-blocking later now that we have a tick service that runs with very low impact on CPU and interrupts.

src/transport_udp.cpp
```diff
@@ -1,237 +1,515 @@
...
+}
+
+bool UDPTransport::send_connect(const TransportConnId conn_id, const Addr& addr) {
+    UdpProtocol::ConnectMsg chdr {};
+
+    chdr.idle_timeout = 120;
+
+    std::lock_guard<std::mutex> _(_socket_write_mutex);
+
+    int numSent = sendto(fd,
```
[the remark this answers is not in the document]
TimEvens: It's to sync both fd_reader and fd_writer threads on writes and updating the shared Connection context.  Not all the methods pass in the connection context right now, but they do pass in the addr.  they main thing that needs to be in sync is the connection context, so some of them can be removed but it's nice to know that sending protocol messages that both fd_writer and fd_reader will do will be in sync on order and when they are writing/updating data.

src/transport_udp.cpp
```diff
@@ -1,237 +1,515 @@
...
+
+    return true;
+}
+
+bool UDPTransport::send_disconnect(const TransportConnId conn_id, const Addr& addr) {
+    UdpProtocol::DisconnectMsg dhdr {};
+
+    std::lock_guard<std::mutex> _(_socket_write_mutex);
+
+    int numSent = sendto(fd,
```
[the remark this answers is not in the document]
TimEvens: Same response as above.

src/transport_udp.cpp
```diff
@@ -1,237 +1,515 @@
...
-  char ip[INET6_ADDRSTRLEN];
+bool UDPTransport::send_keepalive(const TransportConnId conn_id, const Addr &addr) {
+    UdpProtocol::KeepaliveMsg khdr {};
+
+    logger->debug << "conn_id: " << conn_id
+                  << " send KEEPALIVE" << std::flush;
+
+    std::lock_guard<std::mutex> _(_socket_write_mutex);
+
+    int numSent = sendto(fd,
```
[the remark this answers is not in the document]
TimEvens: Same response as above.

include/transport/priority_queue.h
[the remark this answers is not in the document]
TimEvens: I've removed it for now. This change was a fix back before the shutdown with the picoquic transport.  The segfault was caused by the priority queue destructor, but it is a valid question of why was the order not maintained.  I've created issue #112  to work this issue.

src/transport_udp.cpp
```diff
@@ -1,237 +1,515 @@
...
 
-  switch (addr.ss_family) {
-    case AF_INET: {
-      sockaddr_in* s = (sockaddr_in*)&addr;
+            remote.port = s->sin_port;
+            inet_ntop(AF_INET, &s->sin_addr, ip, sizeof(ip));
+            break;
+        }
+        default: {
+            // IPv6
```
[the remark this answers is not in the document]
TimEvens: Updated, but I created issue #110 to fix/add IPv6 support. Right now it doesn't work in UDP, it does in Picoquic.  Not a big deal to add, just want to do that in a different PR.

src/transport_udp.cpp
TimEvens: While not needed, consolidating the send/recv lock to connections lock will result in this being protected by lock as well.

src/transport_udp.cpp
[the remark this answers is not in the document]
paulej: I'm missing the point, I guess. I see no reason to have the mutex locked while calling an OS-level `sendto()` call.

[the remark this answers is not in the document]
TimEvens: You are missing the point.  As mentioned previously it's not about the `sendto()`. The `send_*` methods update the connection context. While a few do not pass in the connection context, the others do.

Maybe your comment is that the ones that do not update it, do not need it... while that is true and it can be removed, it's there right now because it will likely be added back. Regardless, as I mentioned above, I'll remove these in favor of the connection mutex.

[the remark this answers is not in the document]
TimEvens: What I don't like about moving locking out of the `send_*` methods is that they become non thread-safe.  It will require the caller to lock or be super careful. I'll update the method comments to mention that.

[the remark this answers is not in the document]
paulej: In the case, what's to protect? I don't see any reason for the lock.

More generally, I would avoid locking while calling into the OS to send data. Whatever time that takes, the lock is held with no benefit.  I get it that sometimes that's unavoidable.

On locking in every function, while you're right that can be a problem, it can be documented. You don't want the cost of locking and unlocking all the time.

include/transport/priority_queue.h
```diff
@@ -42,6 +42,10 @@ namespace qtransport {
...
+        }
+
```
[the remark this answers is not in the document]
GhostofCookie: Forgot to post review last week! Sorry!

PR #114, opened 2024-01-26
Seattle work week updates

TimEvens:
Tunes UDP pacing, fixes locking, adds latency/one-way trip-time tracking/metric, adds logging when RX and notify queue are full and oldest message is being dropped.

### DIFF
--- a/src/transport_udp_protocol.h
+++ b/src/transport_udp_protocol.h
@@ -1,5 +1,15 @@
 #pragma once
 
+/*
+ * Compiler NOTE:
+ *
+ * __attribute__((__packed__, aligned(1))); is used over "#pragma pack(push, 1)" and "#pragma pack(pop)"
+ *      because only modern compilers support pragma
+ *
+ *  #if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__ is used over "__attribute__((scalar_storage_order("big-endian")))"
+ *      and "#pragma scalar_storage_order big-endian" because only modern compilers support scalar_storage_order.
+ */
+
 namespace qtransport {
     namespace UdpProtocol {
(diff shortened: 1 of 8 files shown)


src/transport_udp_protocol.h
```diff
@@ -67,19 +77,33 @@ namespace qtransport {
...
+            DataMsg() { type = ProtocolType::DATA; }
+
+            struct {
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+                uint8_t discard     : 1;        /// Indicates that data should be discard
+                uint8_t reserved    : 7;        /// Least significant bits reserved
+#else
+                uint8_t reserved    : 7;        /// Least significant bits reserved
+                uint8_t discard     : 1;        /// Indicates that data should be discard
+#endif
```
GhostofCookie, line 102: Since we're using c++20, could instead use constexpr `std::endian::native == std::endian::big` instead, that would be platform agnostic.

GhostofCookie (approved):
Looks good to me

src/transport_udp_protocol.h
TimEvens: That works well for testing during runtime but for the struct in this form it doesn't appear to work. An alternative is:

```
include(TestBigEndian)
TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
set(CXXFLAGS ${CXXFLAGS} -DIS_BIG_ENDIAN=${IS_BIG_ENDIAN})
```

In the CMakeLists.txt, but that may not be as portable as what's in there now. I'll make note of it and if we run into issues, let's try the make method.  If that becomes a problem, we can then change from memcpy style.

PR #123, opened 2024-02-22
Resolve reset and wait issue

TimEvens:
Resolves reset and wait issue where streams reset would not start again. 

Updated reset to mark active immediately a stream after reset. This should speed up iframe objects. 

Fixed issue where reset_and_wait streams would trigger congestion still.

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -1484,6 +1484,11 @@ void PicoQuicTransport::check_conns_for_congestion()
 
         for (auto& [data_ctx_id, data_ctx] : conn_ctx.active_data_contexts) {
 
+            // Skip context that is in reset and wait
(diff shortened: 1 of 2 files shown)


src/transport_picoquic.cpp
```diff
@@ -1814,6 +1834,8 @@ void PicoQuicTransport::mark_stream_active(const TransportConnId conn_id, const
         return;
     }
 
+    data_ctx_it->second.mark_stream_active = false;
```
RichLogan, line 1837: So it's okay to set this to false even if we didn't mark as active because the stream ID = 0?

TimEvens: Yeah, the mark active flag being true prevents duplicate requests to mark active. In this case, we need the next enqueue to mark active again because here we ignore/discard.

PR #127, opened 2024-03-02
Bring the esp changes upto the latest transport (WIP)

suhasHere:
This is rework of PR #63 . This tries to bring the esp additions to match the latest UDP transport.

This PR seems to not break non esp code as confirmed by running echo server and client.

Look forward for some early feedback and would be nice to get this moving along in coming weeks,


suhasHere:
@TimEvens Love to get feedback on this PR. I will be testing this on esp platform in the meanwhile.

cmd/client.cc
```diff
@@ -88,7 +88,7 @@ main()
 {
     char* envVar;
 
-    TransportRemote server = TransportRemote{ "127.0.0.1", 1234, TransportProtocol::QUIC };
+    TransportRemote server = TransportRemote{ "127.0.0.1", 1234, TransportProtocol::UDP };
```
TimEvens, line 91: We should probably make this a config... although I don't think it's urgent because I believe most for testing will modify this code anyway.

cmd/echoServer.cc
```diff
@@ -98,7 +98,7 @@ int main() {
   cantina::LoggerPointer logger = std::make_shared<cantina::Logger>("ECHO");
   Delegate d(logger);
   TransportRemote serverIp =
-      TransportRemote{"127.0.0.1", 1234, TransportProtocol::QUIC};
+      TransportRemote{"127.0.0.1", 1234, TransportProtocol::UDP};
```
TimEvens, line 101: Same as client comment.

TimEvens (approved):
LGTM.  The UDP changes will trigger a merge conflict with some changes that are coming, so it would be good to merge this sooner than later.  Are you planning to add/update more before merge?

suhasHere:
> LGTM. The UDP changes will trigger a merge conflict with some changes that are coming, so it would be good to merge this sooner than later. Are you planning to add/update more before merge?

@TimEvens  yes let's merge it .. I will bring in more changes in small quantum's that needs some more work.

PR #128, opened 2024-03-06
Improve BW detection and support per fragment/object pacing

TimEvens:
Updates the ALG used for bandwidth detection and pacing decrease and increases. 

Adds expired pop/front metrics and logs for messages dropped by the time queue due to TTL expired upon next pop/front. 

Adds per object (includes fragment) delay pacing to smoothen iframe (large object) bursts. Currently it's only applied to UDP and datagram messages with a 30ms spacing.  Will update API to indicate object pacing.

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -549,6 +549,7 @@ PicoQuicTransport::enqueue(const TransportConnId& conn_id,
                            std::vector<qtransport::MethodTraceItem> &&trace,
                            const uint8_t priority,
                            const uint32_t ttl_ms,
+                           const uint32_t delay_ms,
                            const EnqueueFlags flags)
 {
     if (bytes.empty()) {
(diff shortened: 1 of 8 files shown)


suhasHere:
Looks good Tim . Thank you

src/transport_picoquic.cpp
```diff
@@ -620,7 +621,7 @@ PicoQuicTransport::enqueue(const TransportConnId& conn_id,
                       std::move(bytes),
                       std::move(trace)};
 
-        data_ctx_it->second.tx_data->push(std::move(cd), ttl_ms, priority);
+        data_ctx_it->second.tx_data->push(std::move(cd), ttl_ms, priority, delay_ms);
```
suhasHere, line 625: Do we want to make it 0 as well

suhasHere (approved):
One small clarification question, but rest looks good . Thanks

src/transport_picoquic.cpp
TimEvens: thanks for catching that. It should have been zero for this one.

PR #144, opened 2024-05-06
Add qlog, metrics and stream 0 support

### LINKED ISSUE #130: Control stream should use stream Id 0
StreamId 0 is treated special today and control messages are sent on Bidi stream Id 4. We should move to using control stream as stream Id 0, unless it is closed and we move to pick a new one.

TimEvens:
Resolves #130, adds qlog transport config, and below metrics:

* tx_in_transit_bytes -- Number of bytes in transit
* tx_timer_losses (min/max/avg) - Packet losses detected due to timer expiring
* tx_spurious_losses (min/max/avg) - Number of packet lost that were later acked

### DIFF
--- a/src/transport_picoquic.h
+++ b/src/transport_picoquic.h
@@ -132,7 +132,7 @@ class PicoQuicTransport : public ITransport
 
         DataHeader data_header {};                           /// Data header to use for start of stream or for every datagram
 
-        uint64_t current_stream_id {~(uint64_t(0))};         /// Current active stream if the value is >= 4
+        std::optional<uint64_t> current_stream_id;           /// Current active stream if the value is >= 4
 
         uint8_t priority {0};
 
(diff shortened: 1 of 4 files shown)


src/transport_picoquic.h
```diff
@@ -132,7 +132,7 @@ class PicoQuicTransport : public ITransport
 
         DataHeader data_header {};                           /// Data header to use for start of stream or for every datagram
 
-        uint64_t current_stream_id {0};                      /// Current active stream if the value is >= 4
+        uint64_t current_stream_id {~(uint64_t(0))};         /// Current active stream if the value is >= 4
```
suhasHere, line 135: can we make it std::optional<uint64_t> instead ?

TimEvens: Sure, I'll make that change.

PR #147, opened 2024-05-24
Implement stream buffer to support MOQT requirements

### LINKED ISSUE #131: Remove transport headers  when sending data over streams
We add a 4 byte length when using streams to send data. This will work for us internally but will fail any kind of interop.

TimEvens:
MOQT requires per uintvar header parsing, which will have track alias. Track alias is equivalent to data context ID. Unfortunately the transport will not be able to parse this without libquicr (chicken and egg situation). Due to this, MOQT requires a stream of bytes from transport.  Transport looses visibility on mapping which data flow context a new stream belongs to. To mitigate this, libquicr can call `setStreamIdDataCtxId()` to bind a new stream to a data context. This is optional for libquicr. If the binding happens, then metrics will be available at the data context level.  If the mapping does not happen, then metrics will be at the connection level for received data. 

Transmit still has data context and still sends based on libquicr published objects. The main difference is that the transport header does not get added to streams anymore. 

On received notify callback has been changed to use `on_recv_dgram()` and `on_recv_stream()` callbacks. Datagram still has object (e.g., datagram frames) awareness since datagrams require complete objects (complete MOQT message) within each datagram frame. Stream will be delivered based on bytes, which will vary based on several factors. For this reason, a stream buffer of bytes is used to receive bytes on a stream. 

> [!NOTE]
> UDP custom protocol stil has a data header and still works the way it does today. UDP will need to be updated after MOQT QUIC related changes are complete.   

* Resolves #131

### DIFF
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -738,16 +738,10 @@ void PicoQuicTransport::setStreamIdDataCtxId(const TransportConnId conn_id,
                                              DataContextId data_ctx_id,
                                              uint64_t stream_id) {
 
-    logger->debug << "Set data context to stream conn_id: " << conn_id
-                  << " data_ctx_id: " << data_ctx_id
-                  << " stream_id: " << stream_id
-                  << std::flush;
 
     std::lock_guard<std::mutex> _(_state_mutex);
 
-
     const auto conn_it = conn_context.find(conn_id);
-
     if (conn_it == conn_context.end())
         return;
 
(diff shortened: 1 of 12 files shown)


src/transport_picoquic.cpp
```diff
@@ -1600,6 +1371,26 @@ void PicoQuicTransport::emit_metrics()
...
+
+        for (auto& [stream_id, rx_buf]: conn_ctx.rx_stream_buffer) {
+            if (rx_buf.closed && (rx_buf.buf->empty() || rx_buf.checked_once)) {
+                closed_streams.push_back(stream_id);
+            }
+            rx_buf.checked_once = true;
+        }
+
+        for (const auto stream_id: closed_streams) {
+            conn_ctx.rx_stream_buffer.erase(stream_id);
```
GhostofCookie, line 1389: It may be possible to compress this using `std::remove_if`

GhostofCookie (approved):
I think this code looks reasonable to me.

PR #153, opened 2024-07-08
Added python script to get mbedtls python requirements during configure time

BrettRegnier:
Reason:
Need to run a required python package installer for mbedtls for autionation.

Expected:
Run pip install on mbedtls/scripts/basic.requirements.txt

Current:
Doesn't run the script during configure time causing a compile error in PicoTLS and maybe MbedTLS


RichLogan (changes requested):
This doesn't work for me on my mac, because the dependency install fails against the system python. I'm proposing we create a virtualenvironment and install into that: https://github.com/Quicr/transport/pull/154

RichLogan (approved):

PR #154, opened 2024-07-10
Use virtualenv for mbedtls creation

RichLogan:
Create a python virtual env in the build directory (installing mbedtls dependencies) for mbedtls to work inside.


BrettRegnier (approved):
LGTM

PR #156, opened 2024-07-17
Fixup compiler warnings

RichLogan:
- Some returns of temporary variables using `std::move` warn about preventing copy-elision. 
- I believe transport config char pointers should be constant (especially due to string literals being used in some places). 
- Non explicit `this` capture is deprecated (maybe only in clang16, but still). 
- Couple of missing returns in non-void functions.

### DIFF
--- a/include/transport/transport.h
+++ b/include/transport/transport.h
@@ -74,8 +74,8 @@ struct TransportRemote
  */
 struct TransportConfig
 {
-  char *tls_cert_filename;                        /// QUIC TLS certificate to use
-  char *tls_key_filename;                         /// QUIC TLS private key to use
+  const char *tls_cert_filename;                  /// QUIC TLS certificate to use
+  const char *tls_key_filename;                   /// QUIC TLS private key to use
   uint32_t time_queue_init_queue_size {1000};     /// Initial queue size to reserve upfront
   uint32_t time_queue_max_duration {1000};        /// Max duration for the time queue in milliseconds
   uint32_t time_queue_bucket_interval {1};        /// The bucket interval in milliseconds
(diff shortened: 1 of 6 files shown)


src/transport_udp.cpp
```diff
@@ -84,6 +84,7 @@ TransportStatus UDPTransport::status() const {
     } else if (!isServerMode) {
         return clientStatus;
     }
+    return clientStatus;
```
RichLogan, line 87: Wasn't sure about what the correct thing to do in this case is? cc @TimEvens

include/transport/transport.h
```diff
@@ -74,8 +74,8 @@ struct TransportRemote
  */
 struct TransportConfig
 {
-  char *tls_cert_filename;                        /// QUIC TLS certificate to use
-  char *tls_key_filename;                         /// QUIC TLS private key to use
+  const char *tls_cert_filename;                  /// QUIC TLS certificate to use
```
TimEvens, line 77: Challenge is that `TransportConfig` is mutable context/config. const values prevent that.  Changing
this will require implementations to update, such as libquicr. 

I would recommend though that we switch this to string.

include/transport/transport.h
```diff
@@ -74,8 +74,8 @@ struct TransportRemote
...
+  const char *tls_key_filename;                   /// QUIC TLS private key to use
```
TimEvens, line 78: Switch to non-const string.

include/transport/transport.h
```diff
@@ -90,7 +90,7 @@ struct TransportConfig
   uint64_t idle_timeout_ms { 30000 };             /// Idle timeout for transport connection(s) in milliseconds
   bool use_reset_wait_strategy { false };         /// Use Reset and wait strategy for congestion control
   bool use_bbr { true };                          /// Use BBR if true, NewReno if false
-  char* quic_qlog_path;                           /// QUIC LOG file location path, trivially copyable null terminated string
+  const char* quic_qlog_path;                     /// QUIC LOG file location path, trivially copyable null terminated string
```
TimEvens, line 93: Switch to non-const string.

TimEvens (changes requested):
Looks good to me with some updates to transport config to use mutable/non-const string instead.  

The reason transport config is a mutable context, is that it's the context/config we pass down all the way from the app, which some things might need to be updated after construct.  Also, the same config is duplicated and then updated for additional instances, such as changing port, tls cert, ...

src/transport_udp.cpp
TimEvens: Probably can drop

```
else if (!isServerMode) {
        return clientStatus;
    }
```
in favor of having the fallback to just return clientStatus.

RichLogan:
Updated to use `std::string`. I have updated our various usages of this also in the other repos but will wait until we're happy with this.

shenning00 (approved):
Looks good.

PR #157, opened 2024-07-17
Add clang-tidy & initial naming pass

RichLogan:
Adds initial clang-tidy pass, and CI check. In order to not have massive changes, and for my sanity, I'm doing renaming first. Includes on #156 .

The [39b42e3](https://github.com/Quicr/transport/pull/157/commits/39b42e3d6ae04c4bec32a199b2286984bd00c9eb) failure is the one showing the clang-tidy pass correctly failing. 

[bf95adf](https://github.com/Quicr/transport/pull/157/commits/bf95adf509547a5b3c3fc04efecf2f931b40212c) then passes now those violations are fixed.


TimEvens (approved):
Should be good to go, but I didn't see the clang format/tidy changes. Are those coming? 

It will be important to make sure all of our modules follow this... maybe we should have a root repo that acts as a skeleton/base clang policy for all other repos.  This would hopefully keep style/formatting consistent with new and ever changing repos.  

Note, some IDE's have challenges with using the right clang-format/clang-tidy file(s).  It can depend on the project root folder. Many of us work on submodules from some higher base, such as WxQ, new-qmedia, libquicr, ...  It will be important to update all the other repos.

PR #162, opened 2024-08-12
Add note on style/naming/formatting

RichLogan:
This is deliberately intended as a provocative PR to kick off discussion and drive a decision (whatever that may be).


TimEvens (approved):

TimEvens (approved):

RichLogan:
On the call today it sounds like we got agreement here, so I will merge this.

PR #164, opened 2024-08-12
Add & fix clang tidy performance rules

RichLogan:
~(Includes #163 for CI fix)~

Adds `clang-analyzer-*` checks (no changes needed) and the `performance-checks-*` which did have some (minor, but useful if just for consistency) fixes. I will comment them individually.

### DIFF
--- a/include/transport/time_queue.h
+++ b/include/transport/time_queue.h
@@ -228,10 +228,10 @@ namespace qtransport {
 
         TimeQueue() = delete;
         TimeQueue(const TimeQueue&) = default;
-        TimeQueue(TimeQueue&&) = default;
+        TimeQueue(TimeQueue&&) noexcept = default;
 
         TimeQueue& operator=(const TimeQueue&) = default;
-        TimeQueue& operator=(TimeQueue&&) = default;
+        TimeQueue& operator=(TimeQueue&&) noexcept = default;
 
         /**
          * @brief Pushes a new value onto the queue with a time-to-live.
--- a/include/transport/transport.h
+++ b/include/transport/transport.h
@@ -52,7 +52,7 @@ enum class TransportError : uint8_t
 /**
  * Transport Protocol to use
  */
-enum class TransportProtocol
+enum class TransportProtocol: uint8_t
 {
   UDP = 0,
   QUIC
@@ -108,7 +108,7 @@ struct MethodTraceItem {
         delta(0) {
     }
 
-    MethodTraceItem(const std::string method, const time_stamp_us start_time) :
+    MethodTraceItem(const std::string& method, const time_stamp_us start_time) :
             method(method),
             start_time(start_time) {
         delta = (std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now()) -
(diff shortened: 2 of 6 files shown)


include/transport/transport.h
```diff
@@ -52,7 +52,7 @@ enum class TransportError : uint8_t
 /**
  * Transport Protocol to use
  */
-enum class TransportProtocol
+enum class TransportProtocol: uint8_t
```
RichLogan, line 55: Qualify the underlying type to make as small as possible - the enumerations above this already did this 👍🏻

include/transport/transport.h
```diff
@@ -108,7 +108,7 @@ struct MethodTraceItem {
         delta(0) {
     }
 
-    MethodTraceItem(const std::string method, const time_stamp_us start_time) :
+    MethodTraceItem(const std::string& method, const time_stamp_us start_time) :
```
RichLogan, line 111: Probably unintentional copy

src/transport_picoquic.cpp
```diff
@@ -326,7 +326,7 @@ int pq_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode, void
     int ret = 0;
 
     if (transport == NULL) {
-        std::cerr << "picoquic transport was called with NULL transport" << std::endl;
+        std::cerr << "picoquic transport was called with NULL transport" << '\n';
```
RichLogan, line 329: This one and below can go either way. Using '\n' over `std::endl` can be a gain if you don't need to flush the stream, although with `std::cerr` or `std::cout` they always flush anyway, so this isn't actually a gain. 

However, changing it does have the benefit of not ignoring a rule that may make a difference if another stream gets introduced into the codebase later, and using `\n` consistently is the recommendation.

src/transport_picoquic.cpp
```diff
@@ -1336,7 +1336,7 @@ PicoQuicTransport::on_recv_datagram(ConnectionContext* conn_ctx, uint8_t* bytes,
 
     std::vector<uint8_t> data(bytes, bytes + length);
 
-    conn_ctx->dgram_rx_data.push(std::move(data));
+    conn_ctx->dgram_rx_data.push(data);
```
RichLogan, line 1339: This is passed by const reference, so move has no effect.

src/transport_udp.cpp
```diff
@@ -741,7 +741,7 @@ UDPTransport::fd_reader() {
                                       << " port: " << remote.port << std::flush;
 
                         // Notify caller that there is a new connection
-                        _delegate.on_new_connection(_last_conn_id, std::move(remote));
+                        _delegate.on_new_connection(_last_conn_id, remote);
```
RichLogan, line 744: Also passed by const-ref

include/transport/time_queue.h
```diff
@@ -228,10 +228,10 @@ namespace qtransport {
 
         TimeQueue() = delete;
         TimeQueue(const TimeQueue&) = default;
-        TimeQueue(TimeQueue&&) = default;
+        TimeQueue(TimeQueue&&) noexcept = default;
```
RichLogan, line 231: This and the below is [noexcept-move-constructor](https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html):

> Move constructors of all the types used with STL containers, for example, need to be declared noexcept. Otherwise STL will choose copy constructors instead. The same is valid for move assignment operations.

TimEvens (approved):

PR #165, opened 2024-08-14
Replace cantina with spdlog

GhostofCookie:


RichLogan (commented):
I'll defer to @TimEvens for approval but this looks good to me.

Only thing that came up is the difference to the prior dedicated picoquic logger vs. a combined approach here, which I have no opinion on.

PR #167, opened 2024-08-15
Include what you use

RichLogan:
There are a couple of things to call out here:

1. I have done this naively. We may want consumers of this API (and ourselves) to only have to include `<transport/transport.h>` to get all/some subset of the `<transport/*.h>` headers. We can add a pragma to mark those as exported that will be understood if they/us run clang-tidy or IWYU. Perhaps the existing ones transport includes already would make sense(?): cc @TimEvens 

```cpp
#include <transport/uintvar.h>
#include <transport/safe_queue.h>
#include <transport/transport_metrics.h>
#include <transport/stream_buffer.h>
```

2. On MacOS at least, IncludeCleaner suggests some private MacOS headers that contain the actual implementations rather than the more canonical UNIX headers. E.g `timeval` should come from `sys/time.h` not `sys/types/_timeval.h`. IWYU allows a mapping file to correct these, but clang-tidy's version doesn't seem to. I have marked them NOLINT for now.

### DIFF
--- a/src/transport.cpp
+++ b/src/transport.cpp
@@ -3,6 +3,10 @@
 #if not defined(PLATFORM_ESP)
   #include <transport_picoquic.h>
 #endif
+#include <memory>
+#include <stdexcept>
+#include <spdlog/logger.h>
+#include <utility>
 
 namespace qtransport {
 
--- a/src/transport_picoquic.cpp
+++ b/src/transport_picoquic.cpp
@@ -1,27 +1,48 @@
 #include "transport_picoquic.h"
 
+// PicoQuic.
+#include <picoquic.h>
 #include <picoquic_internal.h>
+#include <picoquic_packet_loop.h>
 #include <autoqlog.h>
 #include <picoquic_utils.h>
 #include <spdlog/spdlog.h>
+#include <picoquic_config.h>
+#include <picosocks.h>
 
+// Transport.
+#include "transport/transport.h"
+#include "transport/transport_metrics.h"
+#include "transport/safe_queue.h"
+#include "transport/stream_buffer.h"
+#include "transport/priority_queue.h"
+#include "transport/span.h"
+#include "transport/time_queue.h"
+#include <spdlog/logger.h>
+
+// System.
 #include <arpa/inet.h>
+#include <sys/socket.h>
 #include <cassert>
 #include <cstring>
 #include <ctime>
 #include <iostream>
 #include <thread>
-#include <unistd.h>
 #include <sstream>
+#include <cstdint>
+#include <optional>
+#include <utility>
+#include <memory>
+#include <mutex>
+#include <vector>
+#include <string>
+#include <chrono>
+#include <sstream>
+#include <stdexcept>
 
-#include <arpa/inet.h>
-#include <sys/select.h>
-#include <netdb.h>
 #if defined(__linux__)
 #include <net/ethernet.h>
 #include <netpacket/packet.h>
-#elif defined(__APPLE__)
-#include <net/if_dl.h>
 #endif
 
 #define LOGGER_TRACE(logger, ...) if (logger) SPDLOG_LOGGER_TRACE(logger, __VA_ARGS__)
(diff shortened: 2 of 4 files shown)


src/transport.cpp
```diff
@@ -3,6 +3,10 @@
 #if not defined(PLATFORM_ESP)
   #include <transport_picoquic.h>
 #endif
+#include <memory>
+#include <stdexcept>
+#include <spdlog/logger.h>
```
GhostofCookie, line 8: It preferred this over
```cpp
#include <spdlog/spdlog.h>
```
?

src/transport_picoquic.cpp
```diff
@@ -1,27 +1,48 @@
...
+#include <cstdint>
+#include <optional>
+#include <utility>
+#include <memory>
+#include <mutex>
+#include <vector>
+#include <string>
+#include <chrono>
+#include <sstream>
+#include <stdexcept>
```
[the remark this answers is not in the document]
GhostofCookie: Also, personally (though I'm sure I got this habit from somewhere vaguely official) I like to order the headers as
```cpp
#include "local/header.h"

#include <dependency/header.h>

#include <system/header>
```

That way your dependency headers are always explicitly after your locally declared headers so that you don't run into potential include errors where you didn't include a dependency header in a local header, but that gets hidden when the dependency header comes before the local header in this implementation file. Of course IWYU fixes that issue.

src/transport_picoquic.cpp
```diff
@@ -821,7 +842,7 @@ PicoQuicTransport::ConnectionContext& PicoQuicTransport::createConnContext(picoq
                             /*(const void*)(&((struct sockaddr_in*)addr)->sin_addr),*/
                             conn_ctx.peer_addr_text,
                             sizeof(conn_ctx.peer_addr_text));
-            conn_ctx.peer_port = ntohs(((struct sockaddr_in*)addr)->sin_port);
+            conn_ctx.peer_port = ntohs(((struct sockaddr_in*)addr)->sin_port); // NOLINT (include)
```
GhostofCookie, line 845: What's the linter mad about?

src/transport_udp.cpp
```diff
@@ -1,24 +1,43 @@
...
+#include <sstream>
+#include <stdexcept>
 
+#include <netinet/in.h>
 #include <sys/socket.h>
+#include <sys/select.h> // IWYU pragma: keep
+#include <sys/time.h> // IWYU pragma: keep
 #include <arpa/inet.h>
 #include <netdb.h>
+#include <errno.h>
```
GhostofCookie, line 23: Formatter sort?

src/transport_udp.cpp
```diff
@@ -485,7 +504,7 @@ bool UDPTransport::send_data(ConnectionContext& conn, DataContext& data_ctx, con
  *  - loop reads data from fd_write_queue and writes it to the socket
  */
 void UDPTransport::fd_writer() {
-    timeval to;
+    timeval to; // NOLINT (include).
```
GhostofCookie, line 507: Oh boy, that's a lot of these comments. How do we make it happy?

GhostofCookie (approved):
Comments are optional, I see no reason this shouldn't move forward.

src/transport.cpp
RichLogan: Yeah, it doesn't know about umbrella headers that don't actually define the thing directly, although we can declare them in a mapping file so it does know. Here it didn't seem too horrendous so I didn't bother. But that's the same thing that I'm raising in my comment about `<transport/*.h>`. We can add a pragma in that case if we want.

src/transport_picoquic.cpp
RichLogan: `nthos` is declared in a weird implementation header on Mac at least, so it's asking me to include that rather than the canonical header we want. I can also fix this in a mapping file.

src/transport_udp.cpp
RichLogan: To come in #159

PR #168, opened 2024-08-21
Add CMake module for lint function

RichLogan:
Extract target linting logic to CMake module so it can be easily consumed (and by other quicr projects).


GhostofCookie (approved):
Love it. Ship it.

PR #169, opened 2024-08-21
Rename

RichLogan:
This renames all of `libtransport` to match the project's now declared style guide, and enforces naming via clang-tidy. 

Note that `cmd/` is not currently linted, so I will fix that up subsequently. 

I also added this commit to `.git-blame-ignore-revs`, to mitigate at least some of the pain of having a diff on practically every line. 

I will setup the format harness, and run it, immediately after this PR. 

I don't expect anyone to read this entire diff, but you should be able to reproduce it mostly by using the same `clang-tidy` file and running `clang-apply-replacements clang-tidy-fixes` from the build folder when building with `-DLINT=ON`, and then correcting the API changes in cmd/. 

There was also a name clash on GCC for typedefs matching class names (e.g `using TimeQueue = TimeQueue`). 

Otherwise, nothing functional.


include/transport/priority_queue.h
```diff
@@ -25,8 +25,8 @@ namespace qtransport {
     template<typename DataType, uint8_t PMAX = 32>
     class PriorityQueue
     {
-        using timeType = std::chrono::milliseconds;
-        using timeQueue = TimeQueue<DataType, timeType>;
+        using TimeType = std::chrono::milliseconds;
+        using TimeQueueUs = TimeQueue<DataType, TimeType>;
```
GhostofCookie, line 29: I would instead recommend calling this `TimeQueue_t` or `TimeQueueType` since the time type (which is milliseconds not microseconds) isn't specific when it is passed to it. I.e. `TimeType` has no unit and can change to anything, so the queue type shouldn't have a unit associated with it either.

cmd/client.cc
```diff
@@ -125,23 +125,23 @@ main()
     auto metrics_conn_samples = std::make_shared<SafeQueue<MetricsConnSample>>(10);
     auto metrics_data_samples = std::make_shared<SafeQueue<MetricsDataSample>>(10);
 
-    auto conn_id = client->start(metrics_conn_samples, metrics_data_samples);
+    auto conn_id = client->Start(metrics_conn_samples, metrics_data_samples);
 
-    while (client->status() != TransportStatus::Ready) {
+    while (client->Status() != TransportStatus::kReady) {
```
GhostofCookie, line 130: Oh man, I thought `kReady` was a typo. Google has made some unfortunate choices...

cmd/client.cc
```diff
@@ -174,7 +174,7 @@ main()
         }
 
         // Increase delay if using UDP, need to pace more
-        if (server.proto == TransportProtocol::UDP) {
+        if (server.proto == TransportProtocol::kUdp) {
```
GhostofCookie, line 177: I think it's safe to say readability has been viciously defenestrated.

include/transport/priority_queue.h
```diff
@@ -130,15 +130,15 @@ namespace qtransport {
...
+            std::lock_guard<std::mutex> _(mutex_);
 
-            for (auto& tqueue : _queue) {
-                if (!tqueue || tqueue->empty())
+            for (auto& tqueue : queue_) {
+                if (!tqueue || tqueue->Empty())
                     continue;
 
-                return std::move(tqueue->pop_front());
+                return std::move(tqueue->PopFront());
```
GhostofCookie, line 141: Just realising this `move` is probably bad. But that's a me problem.

include/transport/span.h
```diff
@@ -17,277 +17,277 @@ using Span = std::span<T, N>;
 template<class T, std::size_t Extent = std::numeric_limits<std::size_t>::max()>
```
GhostofCookie, line 17: This file is going to need an exception to the linter: The methods in this **MUST** resemble the **exact** methods of `std::span` otherwise you can no longer easily swap from C++17 to C++20 without breaking a lot. This file is awkward, and its existence is purely because of the C++17 requirement. Weird as it is, it needs to be exempt from the linter.

GhostofCookie (changes requested):

include/transport/stream_buffer.h
```diff
@@ -13,7 +13,7 @@ namespace qtransport {
     template<typename T, class Allocator = std::allocator<T>>
     class StreamBuffer
     {
-        using buffer_t = std::deque<T, Allocator>;
+        using BufferT = std::deque<T, Allocator>;
```
GhostofCookie, line 16: Might have been better to call it `BufferType`.

